PR #172, opened 2025-04-07
chore: update README.md to reflect OpenJII platform details and structure

dominikvrbic:


petar-INFO:
Also maybe its a bit too broad? I think we should remove the "Architecture" and "Repository Structure" sections, IMO those really dont relay a lot of information

petar-INFO (approved):

PR #181, opened 2025-04-24
Add react native BLE basic implementation

ikatun:


petar-INFO:
Looks good in general, though my main concern is not having pnpm as the package manager

https://docs.expo.dev/guides/monorepos/#automatic-configuration writes that since v52, the SDK would automatically configure pnpm? I see that we are using v52, so how come it is not configuring the new project with pnpm?

petar-INFO (approved):

apps/mobile/.prettierrc
dominikvrbic, line 1: This should not be needed. There is a configuration in the root

apps/mobile/eas.json
```diff
@@ -0,0 +1,14 @@
...
+        "buildType": "apk"
+      }
+    },
+    "production": {
+      "android": {
+        "buildType": "app-bundle"
+      }
+    }
+  }
+}
```
dominikvrbic, line 14: From root run pnpm format (after removing the prettier file)

dominikvrbic (approved):

dominikvrbic (commented):
Some small things

PR #214, opened 2025-05-12
feat(web): create experiment

michielschutler:

- [ ] Feature

## Description

Allows users to create a new experiment. It is still very rudimentary and lacks styling.


apps/web/util/schema.ts
```diff
@@ -1,19 +1,18 @@
 import z from "zod";
+import { zExperimentStatus, zExperimentVisibility } from "@repo/api";
 
-export const experimentSchema = z.object({
+export const editExperimentFormSchema = z.object({
```
dominikvrbic, line 4: Please change the names back these are not form schemas but API contractws

michielschutler: Well, actually they are not the same. That is why I changed the names. I cannot directly use the API contracts for the forms.

michielschutler: For example: the edit API contract allows you to only change one field. The form now shows all fields at once and needs those all to be correct.

apps/web/components/edit-experiment.tsx
```diff
@@ -0,0 +1,187 @@
...
+      ...experiment,
+    },
+  });
+
+  async function onSubmit(data: z.infer<typeof editExperimentFormSchema>) {
+    try {
+      // Generate a random userId for demo purposes
+      // In a real app, you would get this from authentication context
+      const userId = "00000000-0000-0000-0000-000000000000";
+
```
dominikvrbic, line 66: Can't you just pass data

michielschutler: Peter wrote this piece also for the other component CreateExperiment. We will replace it as soon as real authentication is in place.

dominikvrbic (approved):

PR #231, opened 2025-05-13
fix: separate versions of react for mobile/web

michielschutler:

- [ ] Bug Fix

## Description

The mobile app and the website need different versions of react for now.


pnpm-workspace.yaml
```diff
@@ -17,6 +17,12 @@ catalog:
 
 catalogs:
   react19:
+    react: ^19.1.0
+    react-dom: ^19.1.0
+    react-hook-form: ^7.56.1
```
dominikvrbic, line 22: probably want to move react-hook-form out of the react sub catalog into a main one,

dominikvrbic (approved):

pnpm-workspace.yaml
michielschutler: Good point. Done!

PR #238, opened 2025-05-13
feat(web): devtools optionally enabled

michielschutler:

- [ ] Feature

## Description

Sometimes it can be practical to have the two development tools enabled on the site, but not always. Having the tools
always enabled during development makes proper styling more difficult.
These will now be disabled by default and can be enabled by creating a local `.env` file containing:

```
NEXT_PUBLIC_ENABLE_DEVTOOLS=true
```

### DIFF
--- a/apps/web/next.config.js
+++ b/apps/web/next.config.js
@@ -1,6 +1,7 @@
 /** @type {import('next').NextConfig} */
 const nextConfig = {
   transpilePackages: ["@repo/ui"],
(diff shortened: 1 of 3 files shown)


petar-INFO:
> Having the tools
always enabled during development makes proper styling more difficult.

How?

michielschutler:
> > Having the tools
> > always enabled during development makes proper styling more difficult.
> 
> How?

Both tools are shown on top of other elements I like to see. E.g. the left tool is shown on top of the user in the sidebar and the right tool is shown on top of toast messages.

petar-INFO (approved):

PR #249, opened 2025-05-19
feat(web): Enhanced create experiment flow

michielschutler:

- [x] Feature

## Description

Enhanced the create experiment flow.


apps/web/app/openjii/experiments/new/page.tsx
```diff
@@ -0,0 +1,29 @@
...
+  return (
+    <AppLayout pageTitle={"New experiment"}>
+        <div className="space-y-6">
+          <div>
+            <h3 className="text-lg font-medium">New experiment</h3>
+            <p className="text-sm text-muted-foreground">
+              Set up a new experiment (project).
+            </p>
+          </div>
+          <NewExperimentForm name={Array.isArray(name) ? name[0] : name} visibilityPrivate={Array.isArray(visibilityPrivate) ? visibilityPrivate[0] === "true" : visibilityPrivate === "true"} />
```
dominikvrbic, line 25: The inline ternaries in <NewExperimentForm> are hard to read. Move this into a variable Generally, logic in JSX is not the best idea

apps/web/components/create-experiment.tsx
```diff
@@ -47,36 +40,10 @@ export function CreateExperiment() {
...
-      toast({
-        description: "Failed to create experiment",
-        variant: "destructive",
-      });
-      console.error("Failed to create experiment:", error);
-    }
+  function onSubmit(data: z.infer<typeof createExperimentFormSchema>) {
+    // Close the dialog and navigate to the new experiment page
+    setOpen(false);
+    router.push(`/openjii/experiments/new?name=${encodeURIComponent(data.name)}&visibilityPrivate=${encodeURIComponent(data.visibilityPrivate)}`);
```
dominikvrbic, line 46: I guess this is just a hack since the endpoint is not hit

apps/web/components/new-experiment.tsx
```diff
@@ -0,0 +1,180 @@
...
+  const router = useRouter();
+  const { mutateAsync: createExperiment, isPending } = useExperimentCreate();
+
+  const form = useForm<z.infer<typeof newExperimentFormSchema>>({
+    resolver: zodResolver(newExperimentFormSchema),
+    defaultValues: {
+      name: name ?? "",
+      description: "",
+      status: zExperimentStatus.enum.provisioning,
+      visibility: visibilityPrivate !== undefined ? (visibilityPrivate ? zExperimentVisibility.enum.private : zExperimentVisibility.enum.public) : zExperimentVisibility.enum.public,
```
dominikvrbic, line 38: Once again, your formatting is not working; also, this does not need to be written like this just set the `visibilityPrivate = false` on line 28,    
 ```
 visibility: visibilityPrivate
        ? zExperimentVisibility.enum.private
        : zExperimentVisibility.enum.public,
```
Also, the status is automatically set on the database level, so there is no need to send it. To create new, I guess it should not even be an option to send it to the backend

File: apps/web/util/schema.ts [hunk 4]
```diff
@@ -1,15 +1,18 @@
 import z from "zod";
 import { zExperimentStatus, zExperimentVisibility } from "@repo/api";
 
-export const editExperimentFormSchema = z.object({
-  id: z.string(),
+export const newExperimentFormSchema = z.object({
```
dominikvrbic, line 4: Why not just use the zCreateExperimentBody? The schemas do not match, now the whole point is not to duplicate them

apps/web/components/new-experiment.tsx
```diff
@@ -0,0 +1,180 @@
...
+            <FormItem>
+              <FormLabel>Status</FormLabel>
+              <Select onValueChange={field.onChange} defaultValue={field.value}>
+                <FormControl>
+                  <SelectTrigger>
+                    <SelectValue placeholder="Select an experiment status" />
+                  </SelectTrigger>
+                </FormControl>
+                <SelectContent>
+                  {(Object.entries(zExperimentStatus.enum)).map(key => {
```
dominikvrbic, line 127: This is for sure not correct. The user should not be able to set it to anything but `archived` or `published`, maybe even not on the API level, the rest are auto-generated statuses

apps/web/components/new-experiment.tsx
```diff
@@ -0,0 +1,180 @@
...
+
+export function NewExperimentForm({ name, visibilityPrivate }: NewExperimentFormProps) {
+  const router = useRouter();
+  const { mutateAsync: createExperiment, isPending } = useExperimentCreate();
+
+  const form = useForm<z.infer<typeof newExperimentFormSchema>>({
+    resolver: zodResolver(newExperimentFormSchema),
+    defaultValues: {
+      name: name ?? "",
+      description: "",
```
dominikvrbic, line 36: No need to set description to an empty string since it's optional on the API level

dominikvrbic (dismissed):

apps/web/components/create-experiment.tsx
michielschutler: It is not meant to be a hack. This is the dialog box shown at the list of experiments. It takes you to the new experiment form with two fields already filled in. This is what I understand from the wireframes.

apps/web/components/new-experiment.tsx
michielschutler: I need to give all fields default values, otherwise I get a warning from React about fields not being controlled.
I will again check my IDE settings for Prettier.

apps/web/components/new-experiment.tsx
michielschutler: I need to give all fields default values, otherwise I get a warning from React about fields not being controlled.

apps/web/components/new-experiment.tsx
michielschutler: Good point. I removed it.

apps/web/app/openjii/experiments/new/page.tsx
michielschutler: Moved the code.

apps/web/util/schema.ts
michielschutler: Good point. I can use it here. I did need to make changes to the edit page (that was already there), since that form cannot reuse the zEditExperimentBody schema. The reason is that in that schema all fields are optional, and this does not apply to the form. I guess that is done because in the API you can choose to only update certain fields. Omitted fields are not updated. Is there a way to use an existing schema and change some optional values into required values?

apps/web/util/schema.ts
```diff
@@ -1,16 +1,16 @@
...
-});
+import { zUpdateExperimentBody } from "@repo/api";
+
+export const editExperimentFormSchema = zUpdateExperimentBody
+  .required({
+    name: true,
+  })
+  .extend({
+    id: z.string().uuid(),
+  });
```
michielschutler, line 11: Made changes here to re-use the existing schema zUpdateExperimentBody.

PR #331, opened 2025-06-05
fix(web): Show error messages for API call

michielschutler:

- [ ] Bug Fix

## Description

If an API call fails, the toast will now show a proper error message.


apps/web/components/new-experiment.tsx
```diff
@@ -46,9 +46,19 @@ export function NewExperimentForm() {
   }
 
   async function onSubmit(data: CreateExperimentBody) {
```
dominikvrbic, line 48: This should not be needed since the query client should do this 
```

// Create a client-side QueryClient instance
const queryClient = new QueryClient({
  defaultOptions: {
    mutations: {
      onError: (error) => {
        toast({
          description: error.message,
          title: error.name,
          variant: "destructive",
        });
      },
    },
  },
});
```
 if that is not working that should be fixed

apps/web/hooks/experiment/useExperimentCreate/useExperimentCreate.ts
```diff
@@ -6,7 +6,11 @@ import { tsr } from "../../../lib/tsr";
  * Hook to create a new experiment
  * @returns Mutation object for creating experiments
  */
-export const useExperimentCreate = () => {
+interface ExperimentCreateProps {
+  onSuccess?: () => unknown;
```
dominikvrbic, line 10: void

dominikvrbic (approved):

PR #370, opened 2025-06-17
feat: refactor experiment member management to support batch addition

vladstoe:

- [x] Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Test Update
- [ ] Other (please describe)

## Description

This PR replaces the old addMember method, which only accepted one member at a time, with a new addMembers method that accepts an array of members. Currently, addMembers is used to add a single member at a time in experiment settings and to add the current user as the admin when an experiment is created.

## Related Issues

- Closes #
- Related to #

## Testing Instructions

- [ ] Test case 1
- [ ] Test case 2

## Changes Made

-
-

## Checklist

- [ ] I have added/updated tests for my changes
- [ ] My code follows the project's style guidelines
- [ ] I have tested these changes locally
- [ ] My changes generate no new warnings
- [ ] I have reviewed my own code


apps/backend/src/experiments/core/models/experiment.model.ts
```diff
@@ -9,12 +10,23 @@ import { experiments } from "@repo/database";
...
+    createdBy: true,
+  })
+  .extend({
+    members: z
+      .array(
+        z.object({
+          userId: z.string(),
+          role: z.enum(["admin", "member"]).optional(),
+        }),
+      )
```
dominikvrbic, line 27: If it is passed it should be at least length of one

apps/backend/src/experiments/application/use-cases/experiment-members/add-experiment-members.ts
```diff
@@ -0,0 +1,103 @@
...
+
+      // Get existing members to perform validations
+      const existingMembersResult =
+        await this.experimentMemberRepository.getMembers(experimentId);
+
+      return existingMembersResult.chain(
+        async (existingMembers: ExperimentMemberDto[]) => {
+          // Check if current user has permission to add members (must be admin or creator)
+          if (
+            experiment.createdBy === currentUserId ||
```
dominikvrbic, line 57: No need to check creator or administrator since by default he is an admin

apps/backend/src/experiments/application/use-cases/experiment-members/add-experiment-members.ts
```diff
@@ -0,0 +1,103 @@
...
+    experimentId: string,
+    members: { userId: string; role?: ExperimentMemberRole }[],
+    currentUserId: string,
+  ): Promise<Result<ExperimentMemberDto[]>> {
+    this.logger.log(
+      `Adding members [${members.map((m) => m.userId).join(", ")}] to experiment ${experimentId} by user ${currentUserId}`,
+    );
+
+    // Check if the experiment exists
+    const experimentResult =
```
dominikvrbic, line 36: Think this could be skipping i am guessing the get members will throw if the experiment does not exits, also it might be a good option to add a utility method on the repository level like isAdmin that takes the user ID and experiment ID and returns boolean

dominikvrbic (changes requested):

PR #405, opened 2025-07-02
feat(mobile): implement redesign and finalize app

ikatun:

- [x] Feature
- [ ] Bug Fix
- [x] Refactor
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Test Update
- [ ] Other (please describe)

## Description

Redesign + refactor


apps/mobile/src/hooks/use-session-store.ts
```diff
@@ -0,0 +1,67 @@
...
+}
+
+interface Actions {
+  setSession: (session: Session) => void;
+  clearSession: () => void;
+}
+
+const SESSION_TOKEN_OVERRIDE = process.env.SESSION_TOKEN_OVERRIDE;
+
+const mockSession: Session | undefined = SESSION_TOKEN_OVERRIDE
```
petar-INFO, line 23: what's this?

apps/mobile/src/protocols/definitions/par.ts
```diff
@@ -0,0 +1,40 @@
...
+    PAR: data.light_intensity,
+    red: data.r,
+    green: data.g,
+    blue: data.b,
+  };
+}
+
+export const label = "Photosynthetically Active Radiation (PAR)";
+
+export const example = {
```
petar-INFO, line 31: What are the examples again?

apps/mobile/src/services/multispeq-communication/mock-device/mock-command-executor.ts
```diff
@@ -0,0 +1,26 @@
+import { protocolsDefinitions } from "~/protocols/definitions";
+import { IMultispeqCommandExecutor } from "~/services/multispeq-communication/multispeq-command-executor";
+import { delay } from "~/utils/delay";
+
+export class MockCommandExecutor implements IMultispeqCommandExecutor {
```
petar-INFO, line 5: What are all the mocks for in the app? (mock experiments etc.)

apps/mobile/src/utils/generate-pseudo-uuid.ts
```diff
@@ -0,0 +1,14 @@
+function simpleHash(str: string): string {
+  let hash = 0;
+  for (let i = 0; i < str.length; i++) {
+    hash = (hash * 31 + str.charCodeAt(i)) | 0;
+  }
+  return Math.abs(hash).toString(16).padStart(8, "0");
+}
+
+export function generatePseudoUUID(obj: any): string {
```
petar-INFO, line 9: https://www.npmjs.com/package/uuid perhaps if it makes life easier?

apps/mobile/src/utils/map-rows-to-measurements.ts
```diff
@@ -0,0 +1,67 @@
+import { generatePseudoUUID } from "~/utils/generate-pseudo-uuid";
+
+export interface MeasurementRecord {
```
petar-INFO, line 3: I guess this is the bronze_data_exp schema? The thing is (and which is why the API doesn't specify what exactly the data schema is), the table schemas can vary. So I can't guarantee this interface will be truthful/correct for long

apps/web/app/api/auth/mobile-redirect/route.ts
```diff
@@ -2,6 +2,12 @@ import type { NextRequest } from "next/server";
 import { NextResponse } from "next/server";
 
 export function GET(request: NextRequest) {
+  // Log all headers
+  console.log("Request Headers:");
+  for (const [key, value] of request.headers.entries()) {
+    console.log(`${key}: ${value}`);
```
petar-INFO, line 8: And especially this one :)

apps/mobile/src/app/(tabs)/profile.tsx
```diff
@@ -0,0 +1,263 @@
...
+      <Card style={styles.infoCard}>
+        <Text
+          style={[
+            styles.infoTitle,
+            {
+              color: theme.isDark ? colors.dark.onSurface : colors.light.onSurface,
+            },
+          ]}
+        >
+          Account Information
```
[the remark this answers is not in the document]
ikatun: Exactly, no internationalization support at the moment.
If needed, we can create a ticket for this, it'd be pretty low effort to add.

apps/mobile/src/protocols/definitions/par.ts
ikatun: Snapshots of real measurements, used for "mock device" mode. Useful to try out the app without actually having the device.

apps/mobile/src/utils/generate-pseudo-uuid.ts
ikatun: Similar, but not the same. This one computes id from hash, not randomly. This way we get the same ID after refetching identical measurement which minimizes rerenders.

apps/mobile/src/utils/map-rows-to-measurements.ts
ikatun: That's OK, this is just the first version.
We'll either support some method of figuring out which table should be used or choose a fixed table for this.

Another possible approach is to simply allow user to choose a table (not sure if that's overly verbose for the user, we can discuss the possibilities for the next version of the app).

apps/mobile/src/hooks/use-session-store.ts
ikatun: no longer needed!

petar-INFO (approved):

PR #506, opened 2025-08-04
chore(database): Setup integration tests to use a different database

michielschutler:

- [x] Other (please describe)

## Description

Separates local development database from test database to prevent conflicts during testing and improve development workflow.

### 🔧 Database Isolation
- **Test Database Configuration**
  - Created dedicated test environment using port 5433 and database `openjii_local_test`
  - Added `.env.test` configuration file for test-specific database settings
  - Updated backend test environment to use isolated test database

- **Environment Management**
  - Introduced `dotenv-cli` for better environment variable handling
  - Created separate script commands for default and test environments
  - Made Docker Compose configuration environment-aware with configurable ports and database names

### 🏗 Script Improvements
- **Dedicated Test Commands**
  - Added `db:test:*` script family for test database operations
  - Updated backend test scripts to use dedicated test setup/teardown commands
  - Maintained backward compatibility for existing development workflows

- **Enhanced Configuration**
  - Made database connection strings dynamic based on environment variables
  - Updated Drizzle configuration to respect environment-specific settings
  - Added Docker Compose project isolation for test containers

### 📁 Files Modified
- Updated backend test configuration and package scripts
- Enhanced database package with environment-specific commands
- Modified Docker Compose and Drizzle configs for multi-environment support
- Added test-specific environment configuration

### 🔌 Integration
- Test runs now use isolated database on port 5433
- Development database remains on port 5432 unchanged
- No migration required - existing workflows continue to work


packages/database/scripts/create-additional-databases.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
```
petar-INFO, line 1: There is no need for any scripts, there should just be a .env.test with a different port for the database. Optionally, the .env.test can also have a different database name, though thats not so important. The docker-compose or the script that creates the container for the test database should then handle passing those env vars to the compose file

petar-INFO (changes requested):

packages/database/scripts/create-additional-databases.sh
michielschutler: I think you misunderstood. The script is included in the Docker container to be run automatically upon startup. See https://hub.docker.com/_/postgres.

petar-INFO: I am aware of init scripts in docker, and I did not misunderstand anything. Maybe you should re-read my comment

michielschutler: Partially. With the proposed changes in this PR, I have it almost running:
- There is one Docker PostgreSQL container which contains two databases: openjii_local and openjii_local_test.
- The environment file .env.test is changed so that is points to the separate test database openjii_local_test.
- When you run an E2E-test, it uses the correct test database.
The only thing missing is that it needs to automatically run the Drizzle Migration against the test database before running the tests.
I don't see how running two separate Docker containers on two different ports would solve that.

dominikvrbic (approved):

PR #553, opened 2025-08-25
feat: MultispeQ Protocol Validation

michielschutler:

- [ ] Feature

## Description

Every time a user adds or edits a MultispeQ protocol, the system validates the JSON-code against a scheme. It will show failing lines in the code editor.

## Changes Made

- Strong validation is introduced for the protocol creation (on API level)
- Appropriate validation error messages are shown in the form, underneath the code editor, on the /platform/protocols/new page
- When you use the API directly, the response will inform you with appropriate status code and error message if the protocol code is invalid
- You are able to (re-)create SPAD, PAR, PHI2 and the UPD14 protocols


packages/api/jest.config.js
petar-INFO, line 1: rebase please, I think this is left over from previous versions

petar-INFO (approved):

PR #818, opened 2025-12-02
fix: consolidates all logouts to one api route

petar-INFO:

- [x] Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Test Update
- [ ] Chore
- [ ] Other (please describe)

## Description

Replaces the `/platform/signout` page with a single API route to handle all logout operations.

### Changes

- **New API Route** (`GET /api/auth/logout`)
  - Centralizes logout logic
  - Redirects to home page with logout confirmation

- **Removed `/platform/signout` Page**
  - Deleted page component and tests
  - No longer needed with API route approach

- **SignOutDialog**
  - Now accepts `open` and `onOpenChange` props
  - Navigates to API route instead of using server action
  - Can be embedded in any component

- **Component Updates**
  - NavUser: Opens dialog instead of navigating to page
  - UnifiedNavbar: Links directly to `/api/auth/logout`
  - DangerZoneCard: Uses new route after account deletion/deactivation

- **Infrastructure**
  - Updated CloudFront cache behavior path from `/*/platform/signout` to `/api/auth/logout`

### Testing

All affected components have updated tests reflecting the new logout flow.

### DIFF
--- a/apps/web/app/api/auth/logout/route.ts
+++ b/apps/web/app/api/auth/logout/route.ts
@@ -1,5 +1,5 @@
 import { handleLogout } from "~/app/actions/auth";
 
 export async function GET() {
-  await handleLogout({ redirectTo: `/?logout=true` });
+  return await handleLogout({ redirectTo: `/?logout=true` });
 }


apps/web/app/api/auth/logout/route.ts
```diff
@@ -0,0 +1,5 @@
+import { handleLogout } from "~/app/actions/auth";
+
+export async function GET() {
+  await handleLogout({ redirectTo: `/?logout=true` });
```
petar-INFO: `?logout=true` because it otherwise won't re-render if you logout on the homepage ("/"), to where we are redirecting to. Tried everything, this is the only thing that worked.

stupid problem, stupid solution 🤷

PR #1009, opened 2026-02-02
feat(infra): Create read and write users for the db

### LINKED ISSUE #954: Dedicated DB writer user
## Description

Currently we use a single, over-privileged pg user to interact with the aurora database. As such we would like to minimise the blast radius by introducing dedicated reader/writer user(s).

Also connection to DB should be done with AWS IAM auth

## Acceptance Criteria

- Database is no longer queried via the fully-privileged admin user

Note: look into aurora db writer / reader endpoints

Blagoj95:

- [x] Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Test Update
- [ ] Chore
- [ ] Other (please describe)

## Description
Added dedicated reader/writer database users to replace the over-privileged admin user in AWS
Create 

## Related Issues

- Closes #
- Related to #954


petar-INFO (commented):
Hm, I thought we should be able to create the postgres user(s) with terraform only, using some postgres provider? Would be nicer I think than typescript and shell scripts.

Its also probably fine if the backend makes use of the writer only, we don't have to make the distinction writer/reader within the backend as if it was compromised somehow, you would have both writer and reader access anyway. So I would just keep "DATABASE" provider (`db`) and make it non-admin and writer

Migrations still need the admin user (for dropping etc.). Unless you make a more privileged writer that has alter, drop etc.

PR #1229, opened 2026-03-31
fix(mobile): send cancel command (-1+) when aborting measurement

dominikvrbic:
* **New Features**
  * Cancel action now explicitly stops active device operations before clearing scan state, making cancellation more reliable.

* **Bug Fixes**
  * Command executor now enforces a single in-flight command and properly rejects superseded or destroyed requests.

* **Tests**
  * Added unit tests covering command execution, cancellation races, destroy behavior, and write failure handling.

* **Chores**
  * Added IoT workspace dependency to the mobile app.

### DIFF
--- a/apps/mobile/package.json
+++ b/apps/mobile/package.json
@@ -30,6 +30,7 @@
     "@react-navigation/native-stack": "^7.14.10",
     "@repo/api": "workspace:*",
     "@repo/auth": "workspace:*",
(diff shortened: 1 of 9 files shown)


petar-INFO (commented):
fix also the title pls (its cut off)

vladstoe (approved):
tested and works

PR #1248, opened 2026-03-31
feat(mobile): show comment indicator in recent measurements (OJD-1443)

dominikvrbic:
## Summary

- Add hasComment indicator icon on measurement list items
- Refactored CompletedState to reuse the shared recent-measurements modal/action flow
- Swap MessageCircle → MessageCircleMore icon throughout
- Tune swipe gesture feel (tighter activation, velocity-first snap)

## Test plan
- [ ] Comment on an unsynced measurement
- [ ] Verify comment icon appears in the recent list
- [ ] Verify no icon for measurements without comments

Closes OJD-1443


apps/mobile/src/components/recent-measurements-screen/recent-measurements-screen.tsx
```diff
@@ -47,6 +48,9 @@ export function RecentMeasurementsScreen() {
       }
       onDelete={() => confirmDelete(item)}
       onSync={item.status === "unsynced" ? () => confirmSync(item) : undefined}
+      hasComment={
+        !!getCommentFromMeasurementResult(item.data.measurementResult as Record<string, unknown>)
+      }
```
xiduzo, line 53: why do we need to type cast here?

`getCommentFromMeasurementResult` should return the correct format?

xiduzo (approved):

apps/mobile/src/components/recent-measurements-screen/recent-measurements-screen.tsx
vladstoe: Nope, measurementResult is typed as object, but getCommentFromMeasurementResult expects Record<string, unknown> as input

PR #1261, opened 2026-04-02
feat(cms, transactional): replace hardcoded email content with dynamic CMS rich text content

vladstoe:

- [x] Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Test Update
- [ ] Chore
- [ ] Other (please describe)

## Description

This PR replaces all hardcoded email templates with content fetched dynamically from Contentful CMS using the rich text editor. 

### `packages/transactional`
- **Replaced all individual email components** with a single generic `email.tsx` that accepts pre-rendered CMS content and a preview string
- **Added `src/lib/contentful.ts`** — fetches emails from Contentful by `internalName` and provides an `interpolate` helper to swap `{{variableName}}` placeholders with runtime values
- **Updated all render functions** to fetch CMS content, interpolate variables, and pass the result into the shared `Email` component
- **Added fallback handling for CMS content** — if Contentful entry is missing or a variable is not provided, safe fallbacks are applied to prevent runtime failures and ensure emails still render with default templates instead of breaking
- **Added `scripts/test-render.ts`** (`pnpm test:render`) — renders all four emails using live Contentful data for local browser preview
- **Config changes** — `tsconfig.json` adds `paths` for `@repo/cms/client` and a project reference to `../cms`; `tsconfig.eslint.json` and `eslint.config.js` are updated so `@typescript-eslint` can resolve cross-package types without affecting the actual build

## `packages/cms`
 
**The core problem:** `@repo/cms` is ESM-only and was built assuming a bundler would handle module resolution. The backend is CJS. When the backend loads `@repo/cms` at runtime, Node's ESM loader requires explicit `.js` extensions on relative imports — without them it throws `ERR_MODULE_NOT_FOUND`.
 
- **Added `.js` extensions** to internal imports in `src/lib/client.ts`, `src/client.ts`, and `src/codegen.ts` — so Node can actually resolve them at runtime across the CJS→ESM boundary
- **Created `src/client.ts` as a slim `@repo/cms/client` subpath** — the main `@repo/cms` barrel pulls in all React feature components (home, about, faq, etc.), which have some files without `.js` extensions and would crash the backend. `client.ts` exports only the Contentful client utilities with no React, no UI components
- **`transactional` now imports from `@repo/cms/client`** instead of `@repo/cms` — avoids loading the entire feature barrel at runtime, which would fail
- **Added `"./client"` to `package.json` exports map** to register the new subpath
- **Added `pageEmails.graphql`** with `ComponentEmailFields` and `PageEmailsFields` fragments (including `internalName`)
- **Added `composite: true` to `tsconfig.json`** to support TypeScript project references
- **Added `tsconfig.eslint.json`** — the regular `tsconfig.json` uses `NodeNext` module resolution (required for CJS interop), but `@typescript-eslint`'s `projectService` can't resolve cross-package types under `NodeNext`. This eslint-only config overrides to `bundler` resolution so type-aware lint rules work, with no effect on compilation or runtime

### `apps/backend`
- Added missing `baseUrl` to the three render function call assertions

## Additional notes

All required changes in [Contentful](https://app.contentful.com/spaces/9h8woqnnje85/entries/1if2vRiFcvfHAqsa57aAvV?previousEntries=) have already been completed.

The **Emails page** and **Email component** both include usage instructions in their *Info* tab, detailing how to configure and use the templates.

## Checklist

Ticket: OJD-1396


.trivyignore
```diff
@@ -9,3 +9,9 @@ CVE-2026-33671
 # NestJS routes are defined at compile time via decorators - no user input reaches route compilation.
 # Will be resolved when Express/router releases an update pinning path-to-regexp@>=8.4.0.
 CVE-2026-4926
+
+# CVE-2026-4800: lodash template() arbitrary code execution - false positive on lodash-es
+# Trivy maps this CVE to lodash-es but lodash-es@4.18.0 does not exist (last release was 4.17.21 in 2021).
+# The vulnerability affects lodash's _.template() with untrusted input; lodash is covered by the
+# pnpm override forcing ^4.18.0. lodash-es is not fixable and the CVE mapping is incorrect.
```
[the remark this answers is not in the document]
vladstoe: Fix suggested by blagoj

dominikvrbic (approved):

dominikvrbic (changes requested):
one sec

packages/transactional/src/scripts/test-render.ts
dominikvrbic, line 1: If this is even needed, make it a simple ts file, no mjs,

packages/transactional/src/lib/contentful.ts
```diff
@@ -0,0 +1,59 @@
...
+): Promise<CmsEmail | null> {
+  const { client } = createContentfulClient();
+  const data = await client.componentEmailByName({ internalName });
+
+  const item = data.componentEmailCollection?.items[0];
+  if (!item?.internalName || !item.preview || !item.content?.json) {
+    console.warn(`[transactional/cms] No CMS email found for "${internalName}".`);
+    return null;
+  }
+
```
xiduzo, line 35: Should we have a fallback / default template (in code)?

packages/transactional/src/render/added-user-notification.ts
```diff
@@ -1,35 +1,58 @@
...
+  preview: string;
 }
 
 export async function renderAddedUserNotification(
   params: RenderAddedUserNotificationParams,
 ): Promise<RenderedEmail> {
-  const { host, experimentName, experimentUrl, actor, role } = params;
+  const { host, baseUrl, experimentName, experimentUrl, actor, role } = params;
+
+  const emailData = await getCmsEmail("added-user-notification", {
```
xiduzo, line 27: Is it documented somewhere which email types the system should have? Also is `getCmsEmail` type safe already (string union) - perhaps we could even use an enum here

packages/transactional/src/render/project-transfer-complete.ts
```diff
@@ -1,27 +1,55 @@
 import { render } from "@react-email/components";
 
-import { ProjectTransferComplete } from "../emails/project-transfer-complete";
+import { Email } from "../emails/email";
+import { getCmsEmail } from "../lib/contentful";
```
xiduzo, line 5: all the `render/*` now are the exact same - except for the `getCmsEmail` parameter. Could this be a generic function now?

packages/transactional/src/scripts/test-render.ts
```diff
@@ -0,0 +1,79 @@
+/**
+ * Renders all emails using the live render functions (which hit Contentful)
+ * and writes the HTML output to /tmp/email-*.html so you can open them in a browser.
+ */
```
xiduzo, line 3: shouldnt this file be `render.test.ts`?

xiduzo (dismissed):

packages/transactional/src/scripts/test-render.ts
vladstoe: nope

.trivyignore
```diff
@@ -9,3 +9,9 @@ CVE-2026-33671
 # NestJS routes are defined at compile time via decorators - no user input reaches route compilation.
 # Will be resolved when Express/router releases an update pinning path-to-regexp@>=8.4.0.
 CVE-2026-4926
+
+# CVE-2026-4800: lodash template() arbitrary code execution - false positive on lodash-es
```
dominikvrbic: update the deps or just remove this trivy is no logner blocking

packages/transactional/src/scripts/test-render.ts
dominikvrbic, line 1: just to understand, this is just a way to create an HTML file, also, this open file will not work in CI, but overall i dont get hte point of this file

vladstoe: It's just a helper to visualize the emails with cms content loaded without having to run the local server and do the required action to see it in mailcatcher for example. I can remove it if you think it is useless

petar-INFO: In theory its not a bad script, but in practice it reads like AI slop a bit. So either just polish the script a little or remove it

- tmp dir?
- ascii arrow
- hardcoded env configuration

petar-INFO (approved):

dominikvrbic (approved):
lgtm

PR #1307, opened 2026-04-16
refactor(web): overhauls testing approach & principles

petar-INFO:

- [ ] Feature
- [ ] Bug Fix
- [x] Refactor
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Test Update
- [ ] Chore
- [ ] Other (please describe)

## Description

Overhauls the frontend test infrastructure and rewrites most existing tests in `apps/web` to use a shared, consistent foundation.

The old setup had each test file independently importing `@testing-library/jest-dom`, setting up its own mocks for `next/navigation`, `@repo/i18n`, `better-auth`, and jsdom polyfills. This PR centralizes all of that into a proper test harness:

- `test/setup.ts` - global jsdom polyfills (ResizeObserver, matchMedia, pointer capture), RTL cleanup, and shared module mocks for i18n, next/navigation, better-auth, and PostHog.
- `test/test-utils.tsx` - custom `render` and `renderHook` wrappers that provide QueryClient and tsr providers, plus a `renderWithForm` helper for components using react-hook-form context. Tests import from `@/test/test-utils` instead of `@testing-library/react` directly.
- `test/msw/` - MSW server setup with a `server.mount()` helper that takes a ts-rest contract endpoint and returns a request spy, making API mocking declarative and consistent.
- `test/factories.ts` - typed factory functions for creating test data (experiments, macros, protocols, users, visualizations, etc.) with sensible defaults and partial overrides.
- `.env.test` - test-specific env vars, tracked in git via a `.gitignore` exception.

The vitest config now points to the centralized setup file and drops the inline env vars. The eslint config disables `@typescript-eslint/unbound-method` in test files. `packages/ui` also gets a shared ResizeObserver polyfill in its vitest setup.

Across ~287 test files, tests are rewritten to use the shared utilities - removing per-file boilerplate, using factories for test data, using `server.mount()` for API mocking, and using accessibility-based queries (`getByRole`, `getByLabelText`) over `data-testid` where possible.

Production code changes are minor, surfaced while making components testable:

- `experiment-archive` / `experiment-delete` - wrap `mutateAsync` calls in try/catch (the promise rejects even when `onError` handles it)
- `iot-protocol-runner` - add a ref guard to prevent double-runs during async execution
- `question-card` - call `onToggleRequired` via arrow function to avoid passing the event object
- `unified-navbar` / chart renderers - fix i18n imports from `react-i18next` to `@repo/i18n` / `@repo/i18n/client`
- `useExperimentDataUpload` - remove unused `data` parameter and console.log
- `packages/api` - export the `ExperimentAccess` type (needed by test factories)
- `apps/backend/fp-utils` - formatting only


apps/web/TESTING.md
```diff
@@ -0,0 +1,184 @@
+# Testing — `apps/web`
```
petar-INFO, line 1: fellow devs @xiduzo @vladstoe @dominikvrbic , see how you like this. 

let me know if you would like something different and/or additional to this.

if you are OK with this, then assume all* tests in the diff are done exactly as described here, i.e. (obviously) don't review the 50k line diff

*possibly I missed some, its 300 tests 🤷

dominikvrbic (approved):
Lgtm

apps/web/app/[locale]/(auth)/login-error/page.test.tsx
```diff
@@ -1,119 +1,55 @@
...
-    <div data-testid="auth-hero-section">Hero {locale}</div>
-  ),
+  AuthHeroSection: () => <section aria-label="auth hero" />,
 }));
-
 vi.mock("~/components/auth/error-content", () => ({
-  ErrorContent: ({ locale }: { locale: string; error?: string; errorDescription?: string }) => (
-    <div data-testid="error-content">Error content - {locale}</div>
+  ErrorContent: ({ error }: { error?: string }) => (
+    <section aria-label="error content">{error}</section>
```
Comment (xiduzo) [inline apps/web/app/[locale]/(auth)/login-error/page.test.tsx:16]: yeah! F*ck `data-testid`s 🚀

apps/web/app/[locale]/(auth)/login/page.test.tsx
```diff
@@ -1,130 +1,55 @@
...
-globalThis.React = React;
-
-// --- Mocks ---
-const mockAuth = vi.fn();
-vi.mock("~/app/actions/auth", () => ({
-  auth: (): unknown => mockAuth(),
-  providerMap: [],
-}));
-
 vi.mock("@/components/navigation/unified-navbar/unified-navbar", () => ({
```
Comment (xiduzo) [inline apps/web/app/[locale]/(auth)/login/page.test.tsx:8]: It's the second time I've seen this setup.

In `jest` there is this concept of `__mock__` folders that will do this repetitive work for you (https://jestjs.io/docs/manual-mocks#mocking-user-modules)

Perhaps we can also think about some global mocking patterns - potentially as a next step and only _unmock_ where needed.

apps/web/components/protocol-settings/protocol-details-card.test.tsx
```diff
@@ -217,16 +198,24 @@ describe("ProtocolDetailsCard", () => {
...
   });
 
-  it("should show loading state during update", () => {
-    useProtocolUpdateMock.mockReturnValue({
-      mutateAsync: vi.fn(),
-      isPending: true,
+  it("should show loading state during update", async () => {
+    server.mount(contract.protocols.updateProtocol, {
+      body: createProtocol(),
+      delay: 999_999,
```
xiduzo, line 204: I might have missed where this was

What does `server.mount` do? Is this equivilant to `.spyOn(contract.protocols.updateProtocol, ....)`?

apps/web/components/protocol-settings/protocol-settings.test.tsx
```diff
@@ -79,55 +56,76 @@ describe("ProtocolSettings", () => {
...
 
-  it("should show not found state when data is missing", () => {
-    useProtocolMock.mockReturnValue({
-      data: undefined,
-      isLoading: false,
+  it("should show not found state when data is missing", async () => {
+    server.mount(contract.protocols.getProtocol, {
+      body: { message: "Not found" } as never,
+      status: 404,
     });
```
xiduzo, line 63: I did not open the code myself, but I assume that this will do type inference? E.g. when the `contract.protocols.getProtocol` can return either a `404` or `200` based on the contract the status can only be either `404` or `200`?

apps/web/components/side-panel-flow/__tests__/analysis-panel-flow.test.tsx
```diff
@@ -1,96 +1,50 @@
...
-        }),
-      },
-    },
-  },
-}));
-
 describe("<AnalysisPanel />", () => {
   beforeEach(() => {
-    vi.clearAllMocks();
+    server.use(
```
xiduzo, line 38: Similar to the `server.mount` questions -> this is equivalent to the `.spyOn().mockReturn` or `.spyOn().mockImplementation`

apps/web/components/protocol-settings/protocol-details-card.test.tsx
petar-INFO: Http interceptor, like nock

apps/web/components/side-panel-flow/__tests__/analysis-panel-flow.test.tsx
petar-INFO: Http interceptor, like nock

apps/web/components/protocol-settings/protocol-settings.test.tsx
petar-INFO: No unfortunately not, we would have to hand roll that 

(property) MountOptions.body?: unknown
JSON response body. Omit for 204 or error defaults

(property) MountOptions.status?: number | undefined
HTTP status. Auto-detected from method if omitted (200/201/204). ≥400 = error.

petar-INFO: also i dont think we need that but down to talk about it

apps/web/app/[locale]/(auth)/login/page.test.tsx
Comment (petar-INFO) [reply apps/web/app/[locale]/(auth)/login/page.test.tsx:8]: Yeah a lot of pages mock it, because page level tests don't necessarily care about the child components

It appears in 5 files I don't think it warrants a global mock and in general you will "have to"/will want to mock components in pages anyway

also we don't use jest but i get your point 👍

apps/web/components/protocol-settings/protocol-details-card.test.tsx
petar-INFO: bit more info;

msw-based , but I rolled out some utility so it is a bit more convenient to setup interceptors on demand and on test level .. otherwise its intended to work through global handlers mainly

https://mswjs.io/docs/best-practices/structuring-handlers/

apps/web/components/protocol-settings/protocol-settings.test.tsx
xiduzo: I wrote some ts-util that added this for the FE, check https://github.com/Jan-IngenHousz-Institute/open-jii/blob/main/apps/web/lib/tsr.ts#L68. Perhaps that can be (re)used

xiduzo: Implementation example at https://github.com/Jan-IngenHousz-Institute/open-jii/blob/main/apps/web/hooks/macro/useMacroCreate/useMacroCreate.ts#L33

petar-INFO: I've added something along the lines what you describe here https://github.com/Jan-IngenHousz-Institute/open-jii/pull/1307/changes/d918508fe15594403b036bc5e389aa7fa97cf90c

PR #1310, opened 2026-04-17
fix(mobile): debug-friendly raw measurements (no gzip, copy JSON)

dominikvrbic:
## Summary
- Stop gzipping measurement results before writing to SQLite. Rows are now stored as plain JSON strings; the `gz:`-prefix-aware reader is kept so pending uploads from older installs still decode.
- Add a **Copy JSON** button to the Raw data tab in the measurement preview (uses `expo-clipboard`). Makes it easy to grab the raw payload off-device when debugging why the phone and cloud macros produce different results on the same measurement.

MQTT sample compression is unchanged — the wire payload is still gzipped.

## Test plan
- [ ] Take a measurement → raw data tab shows a "Copy JSON" button; tapping it shows a toast and pastes valid JSON elsewhere.
- [ ] Fresh install: save a measurement, kill the app, reopen → measurement still visible in Recent.
- [ ] Upgrade path: install the previous build, create a failed upload, upgrade to this build → the legacy gzipped row still decodes and uploads.
- [ ] Verify MQTT payload still has `_sample_encoding: "gzip+base64"` (cloud side continues to ungzip).


dominikvrbic:
Superseded by #1320 — the negative-phi2 fix (OJD-1463) lives there on a clean branch off main.

PR #1337, opened 2026-04-22
fix(mobile, web): update auto-proceeded summary to display only the first answer

vladstoe:
Refs: OJD-1435

- [ ] Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Test Update
- [ ] Chore
- [ ] Other (please describe)

## Description

This PR updates AutoProceededSummary banner to only display the answer for the first auto-proceeded question. Previously, the banner rendered all auto-proceeded answers.

## Checklist

- [ ] I have added/updated tests for my changes

### DIFF
--- a/apps/mobile/src/screens/measurement-flow-screen/components/flow-nodes/question-node/auto-proceeded-summary.tsx
+++ b/apps/mobile/src/screens/measurement-flow-screen/components/flow-nodes/question-node/auto-proceeded-summary.tsx
@@ -45,14 +45,16 @@ export function AutoProceededSummary({ currentNodeId, iterationCount }: AutoProc
 
   const firstManualNodeId = getCachedFirstManualNodeId(iterationCount);
 
-  const autoProceededWithAnswers = flowNodes
-    .filter((n) => n.type === "question" && isAutoincrementEnabled(n.id))
-    .filter((n) => !!getAnswer(iterationCount, n.id)?.trim());
+  const firstAutoProceededWithPlot = flowNodes.find(
+    (n) =>
+      n.type === "question" &&
+      isAutoincrementEnabled(n.id) &&
+      !!getAnswer(iterationCount, n.id)?.trim() &&
+      /(plot)/i.test(`${n.name ?? ""} ${n.content?.text ?? ""}`),
+  );
 
   const show =
-    iterationCount > 0 &&
-    autoProceededWithAnswers.length > 0 &&
-    firstManualNodeId === currentNodeId;
+    iterationCount > 0 && !!firstAutoProceededWithPlot && firstManualNodeId === currentNodeId;
 
   if (!show) return null;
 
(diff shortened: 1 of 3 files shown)


apps/mobile/src/screens/measurement-flow-screen/components/flow-nodes/question-node/auto-proceeded-summary.tsx
```diff
@@ -45,14 +45,16 @@ export function AutoProceededSummary({ currentNodeId, iterationCount }: AutoProc
 
   const firstManualNodeId = getCachedFirstManualNodeId(iterationCount);
 
-  const autoProceededWithAnswers = flowNodes
-    .filter((n) => n.type === "question" && isAutoincrementEnabled(n.id))
-    .filter((n) => !!getAnswer(iterationCount, n.id)?.trim());
+  const firstAutoProceededWithPlot = flowNodes.find(
```
dominikvrbic: Don't think this is a good idea, what if they call it plant or anything else or in another language  then it will not work

dominikvrbic (changes requested):

apps/mobile/src/screens/measurement-flow-screen/components/flow-nodes/question-node/auto-proceeded-summary.tsx
vladstoe: I had it made like this until Romy got an answer to this specific question actually, if they usually name the questions plot or not. I'll remove the check based on what they say. Or can you confirm already?👀

dominikvrbic: Honestly, I wouldn't implement this in this way at all, even if Romy confirms "plot" is a common naming. I pulled every PhotosynQ question I could find (**45k questions across \~19.7k projects**) and did a semantic analysis on the identifier concepts people actually use. The data doesn't support hardcoding "plot":

* **"plot" is only the 6th most common identifier.** Ranked by project reach: `treatment` (5,592 projects), `rep` (4,074), `plant` (3,783), `leaf` (3,265), `genotype` (2,104), `plot` (1,729). So this banner silently won't render for the majority of experiments.
* **291 distinct label variants for "plot" alone** across those 1,729 projects (`Plot`, `N PLOT`, `Plot no`, `parcela`, `parcelle`, `Parzelle`, `xiao qu`, …). The current `/plot/i` catches the English ones and misses the Spanish / French / German / Chinese equivalents entirely.
* Only **62.8%** of labels match *any* English keyword concept — 37% are free-form (`"Which treatment was applied?"`, `"¿Cuál es la variedad?"`), which this approach can't handle at all.

On top of the data:

* **It's guesswork / fragile.** Any typo, translation, or unrelated use ("replot count") breaks it. Relying on a naming convention that isn't enforced anywhere isn't reliable.
* **It's undocumented/invisible.** If you don't know this regex exists, the UX silently changes depending on a substring match. Confusing for anyone touching this code later — including us in 6 months.
* **Edge cases aren't defined.** What if a plot has multiple plants? What if there are several auto-proceed fields worth surfacing? We pick the first match and hide the rest — that might be wrong depending on the experiment.
* **The banner header "Your current plot" is hardcoded** (line 74), so this whole feature is already plot-specific — which is exactly the problem: **this is not only a JII platform; features need to be generic.** Baking "plot" into mobile string-matching + copy assumes every experiment on this platform revolves around plots, which the data shows isn't true even for JII's sister community (PhotosynQ).

Reminder for everyone on this: we shouldn't be hardcoding JII-specific terminology into the generic mobile flow. We could feature-flag for JII users and document it, but that's also not great — we'd ship a divergent UX that other tenants can't turn on cleanly.

I'd rather wait until we have proper typed question identifiers (or a "display in summary" flag on the question itself) and key off that.

apps/web/components/question-card/question-card.tsx
```diff
@@ -62,6 +62,13 @@ export function QuestionCard({
           />
         </div>
 
+        {/* Plot / ID question hint */}
+        <p className="from-jii-medium-green/10 to-jii-dark-green/10 text-jii-dark-green mb-6 rounded-lg bg-gradient-to-r px-3 py-2 text-xs">
+          Note: If this is a <strong>plant identifier</strong> question, place it{" "}
+          <strong>first</strong> in the flow. The mobile app will show it as a confirmation banner
+          throughout the iteration.
+        </p>
```
xiduzo, line 70: This is not really foolproof - "the first one will be shown as the banner" 🤔

vladstoe: It was discussed with Romy and JII, and this works for their use case for now as a temporary fix

dominikvrbic: We are not doing it like this. I will close this pr

PR #1407, opened 2026-05-06
fix(web/workbook): output cell + protocol cell polish, run timer, driver timeout

dominikvrbic:
## Summary

Polish pass on the workbook notebook cells uncovered while testing.

### Output cell ([OJD-593](https://linear.app/openjii/issue/OJD-593))

- Table view rendered as JSON for the typical protocol/macro output shape (a single object). It now renders a 2-column key/value table, with array-of-objects fields rendered recursively as sub-tables.
- Output block had no collapse/expand affordance even though the schema already carries `isCollapsed`. Added a chevron toggle next to the OUTPUT label that hides the body while keeping the toolbar visible.
- JSON tab gained a floating Copy button (top-right, matches the pattern used elsewhere) that swaps to a check icon for ~2 s on success.
- JSON-viewer height capped (`max-h-[480px]` with overflow-auto) so long payloads scroll instead of bloating the cell.
- Numeric-array fields render inline sparklines in the table; click expands a full Plotly line chart below.
- Strings translated, output-cell module split into `output-cell.tsx` / `output-cell-charts.tsx` / `output-cell-render-data.tsx` for better isolation.

### Protocol cell ([OJD-590](https://linear.app/openjii/issue/OJD-590))

- The editor was hard-coded read-only. Now wired up the same way as the macro cell: `useSession` + ownership check against `protocolData.body.createdBy`, `useProtocolUpdate` with a 1 s debounced auto-save, and an `onChange` that only persists when the editor contents parse as a JSON array (invalid JSON stays local).
- Newly-created protocols (`code: []`) used to render "Could not load protocol code", which made them un-editable. Now `[]` renders an empty editor; the error message is reserved for genuine fetch failures (`protocolData == null`).

### Live run timer

`CellWrapper` now renders a small live `0ms / 1.2s / 16s` counter next to the spinner while `executionStatus === "running"`. Mounts on the running transition, ticks every 100 ms, and unmounts when the status changes. Format breakpoints (ms under 1 s, x.x s for 1–10 s, integer seconds beyond) keep the badge a stable width.

### Driver timeout

`MULTISPEQ_FRAMING.DEFAULT_TIMEOUT` raised from 30 s → 10 min. Long protocols (extended dark adaptation, full PIRK/DIRK runs) can take several minutes per command; the 30 s default was tripping mid-run on those. 10 min still surfaces a hung device as an error eventually.

## Test plan

- [x] `pnpm vitest run components/workbook` — 195/195 pass.
- [x] `pnpm vitest run packages/iot/src/driver/multispeq` — 16/16 pass (driver tests pick up the new default timeout).
- [x] Lint + Prettier clean across all changed files.
- [x] Manual: create a fresh protocol, confirm the cell drops you into an empty editable editor; edit + auto-save round-trip works.
- [x] Manual: run a protocol and confirm the Output cell's Table tab now shows a key/value table for the device payload.
- [x] Manual: collapse + expand an Output cell; verify the chevron + body visibility.
- [x] Manual: switch to JSON view and click Copy; verify the icon swaps and clipboard contains the payload.
- [x] Manual: run any protocol and watch the run timer tick next to the spinner; verify it stops on completion.

Closes OJD-593
Closes OJD-590

The MultispeQ timeseries view that was on this branch earlier has been reverted — needs more iteration on the wall-clock anchoring / autogain visualisation before it ships. Will go in its own PR. The TypeScript port of `protocol_pipeline.pipeline` and the bundled UNZA_15 fixtures stay on `jii-data-platform` `main` (committed separately) for the python pipeline alignment.


vladstoe (approved):
test plan passes, 

also probably not part of the stories but, for the protocol cell while running, a stop button ⏹️ instead of the play button that sends the cancel command would be nice

PR #1423, opened 2026-05-07
revert: mobile macro wrapper contract (4be4d77)

dominikvrbic:
Reverts #4be4d77 (`fix(mobile): align macro contract with Databricks/sandbox by passing full wrapper to runner`).

The commit broke macro processing in mobile 1.20: legacy macros that call `GetProtocolByLabel` / read `json.set` (i.e. effectively all of them, via `math.lib.js.txt`) blow up with `Cannot read property 'filter' of undefined` because `json` is now the wrapper `{sample: [...], device_id, ...}` and has no `.set`.

The commit's premise — that Databricks/sandbox pass the wrapper to the macro — doesn't hold. On the online path:

- `apps/data/src/pipelines/centrum_pipeline.py:558` does `parse_json(sample)`, so the `data` column the UDF sends is the **sample array**, not the wrapper.
- `apps/macro-sandbox/lib/wrappers/wrapper.js`'s `unwrapMeasurement` then unwraps that array to `[0]`.
- Net result: `json` is a single sample with `.set` directly on it.

So the real contract — same in both runtimes — is `json = single sample`. Mobile's pre-`4be4d77` code already matched this (per-sample loop, `MacroOutput[]`). The revert restores that.

Reported in Slack: phone fails to process measurements during student induction; openJII processes the same measurement fine.

## Test plan

- [ ] Verify Tom's failing measurement (14:54 in experiment d2b9c92a-4695-4017-bc3d-78924886e0a2) processes on the phone
- [ ] Spot-check a couple of other macros that use `GetProtocolByLabel`

### DIFF
--- a/apps/mobile/src/components/measurement-result/measurement-result.tsx
+++ b/apps/mobile/src/components/measurement-result/measurement-result.tsx
@@ -41,13 +41,10 @@ export function MeasurementResult({
     return await applyMacro(rawMeasurement, macro);
   }, [rawMeasurement, macro]);
 
-  const messageGroups: MacroMessageGroup[] = processedMeasurement?.messages
-    ? [processedMeasurement.messages]
-    : [];
-
-  const processedKeys = processedMeasurement
-    ? Object.keys(processedMeasurement).filter((key) => key !== "messages")
-    : [];
+  const messageGroups: MacroMessageGroup[] =
+    processedMeasurement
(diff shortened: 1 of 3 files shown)


vladstoe (approved):
Tried with protocols PAR Only 2.0 and UNZA_PIRK_DIRK_LightPotential_14 and analysis is also displayed in the app now, compared to before with same flows.

PR #1446, opened 2026-05-12
feat(iot): support large IoT payloads via S3 pre-signed upload URLs

### LINKED ISSUE #1303: Support IoT payloads over 128 KB
## Story

This work provides a supported path for IoT payloads that exceed AWS IoT Core’s 128 KB MQTT limit. It introduces an S3-based upload flow for both authenticated clients and X.509-authenticated devices, so large measurements can still enter the platform and continue through the existing processing flow.

Some current and future sensor scenarios can exceed the MQTT payload limit, including full spectral dumps, offline batch uploads, and dense high-frequency sampling sessions. Without an alternative transport, these payloads cannot be ingested reliably. This story adds a backend-issued pre-signed upload path for session-authenticated clients and a credential-provider path for certificate-authenticated devices.

### Core problem it solves

IoT Core cannot accept payloads larger than 128 KB, and the platform currently has no supported alternative path for large device data.

### Outcome / success

* Authenticated clients can request an upload URL and send large payloads directly to S3.
* X.509-authenticated devices can obtain temporary credentials and upload directly to S3.
* Uploaded large payloads land in a controlled S3 prefix with the metadata needed for downstream processing.
* The upload path enforces size and validation constraints.
* Large payloads can continue into the same downstream data pipeline as standard MQTT traffic.

## Acceptance Criteria

* A new backend capability exists to generate upload URLs for large IoT payloads.
* Gateway and IoT service routing support the new upload-url flow.
* Request and response contracts are defined and validated.
* Upload requests enforce size limits and reject invalid parameters.
* IoT Core Credential Provider infrastructure exists for X.509-authenticated devices.
* Devices using X.509 certificates can obtain temporary IAM credentials scoped to the correct S3 prefix.
* S3 event-driven processing is configured for large-payload uploads.
* Uploaded objects can be processed into the existing platform ingestion path, or into the agreed direct S3-based path if that replaces the passthrough step.
* Infrastructure, documentation, and monitoring cover both client upload flows.

## Dependencies & Risks

* Depends on the raw-data bucket from the S3 archive work.
* Depends on the extracted IoT service and gateway transport path.
* Requires careful coordination between backend contracts, S3 key structure, and downstream ingestion.
* Large-payload handling adds complexity across S3, IAM, event notifications, and processing infrastructure.
* The final processing path should stay aligned with the Databricks ingestion approach to avoid temporary duplication.

## Additional Context

There are two supported upload models:

* Session-authenticated clients request a pre-signed URL from the backend and upload directly to S3.
* X.509-authenticated devices use IoT Core Credential Provider to obtain temporary IAM credentials and upload directly to S3.

The intent is to keep the data format consistent with the existing IoT ingestion channel so transport changes do not force schema changes downstream.

Blagoj95:

- [x] Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Test Update
- [ ] Chore
- [ ] Other (please describe)

## Description

IoT Core rejects MQTT messages larger than 128 KB. This adds a
`POST /api/v1/iot/upload-url` endpoint that returns a pre-signed S3
PutObject URL so clients can bypass the MQTT size limit by uploading
large payloads directly to S3.

- **Backend** — new `AwsS3Service` generates pre-signed URLs (15 min
  expiry) for `large-iot/{experimentId}/{uuid}.json` on the IoT archive
  bucket; follows the same port/use-case/adapter pattern as the existing
  IoT credentials flow
- **API contract** — added `POST /api/v1/iot/upload-url` with Zod
  schemas for request (`experimentId`) and response
  (`uploadUrl`, `key`, `expiresAt`)
- **Infrastructure** — `backend_iot_s3_presign` IAM policy added to the
  `iot-core` module and attached to the ECS task role in all three
  environments (dev/prod/dr); `AWS_IOT_ARCHIVE_BUCKET_NAME` wired as an
  env var on the backend ECS service

## Related Issues

- Closes #
- Related to #


Blagoj95:
@coderabbitai review

apps/web/components/auth/registration-form.test.tsx
```diff
@@ -284,18 +284,23 @@ describe("RegistrationForm", () => {
...
     await user.type(screen.getByLabelText("registration.firstName"), "Multi");
     await user.type(screen.getByLabelText("registration.lastName"), "Submit");
-
     await user.click(screen.getByRole("checkbox"));
+
     const submitButton = screen.getByRole("button", { name: "registration.register" });
 
-    // First click
-    await user.click(submitButton);
+    // Use fireEvent (no internal act() wrapper) so the 100ms setTimeout in the mock
```
petar-INFO: What are these additions?

apps/backend/src/common/modules/aws/services/s3/s3.service.spec.ts
```diff
@@ -0,0 +1,66 @@
+import type { PutObjectCommand } from "@aws-sdk/client-s3";
```
petar-INFO, line 1: use here that aws client mock if possible for the test suite

petar-INFO (approved):

apps/backend/src/common/modules/aws/services/s3/s3.service.spec.ts
Blagoj95: `getSignedUrl` bypasses `client.send()` so `mockClient` can't intercept it, and `vi.mock` doesn't replace the ESM binding in the service scope — `createSignedUrl` is the intentional test seam for this.

apps/web/components/auth/registration-form.test.tsx
Blagoj95: I guess added those when resolving merge conflict. Now this file is same as `main` as it's not related to changes done here

petar-INFO (approved):

PR #1454, opened 2026-05-13
fix(mobile): offline measurements bug + Cognito burst fix + pending status

dominikvrbic:
## Summary

Three connected fixes for the bug Tom reported after the Grebbedijk field session (~2000 measurements/day), plus a small CI gate added at the end so the next typecheck regression doesn't slip through.

- **OJD-1514** — Offline measurements vanished from Recent until reconnect, by which point they had synced and the flag UI was gone.
- **OJD-1515** — Post-field upload burst saturated AWS Cognito; the Identity Pool had accumulated ~20K junk identities and credentials were re-fetched on every publish.
- **OJD-1516** — New `pending` measurement status (distinct from `failed`) + DB-level CHECK constraint via a SQLite migration.

Linked tickets:
- https://linear.app/openjii/issue/OJD-1514
- https://linear.app/openjii/issue/OJD-1515
- https://linear.app/openjii/issue/OJD-1516

## OJD-1514 — Save-before-upload

`use-measurement-upload.ts` and `use-questions-upload.ts` now persist the measurement to local SQLite **as `pending`** *before* the MQTT publish. On success → `markUploaded` flips to `successful`; on failure → `markFailed` flips to `failed`. The row appears in Recent immediately, regardless of network state, and the flag UI is reachable until the upload confirms.

The publish path and the local-state write are in separate try/catch blocks: a `markUploaded` failure that fires *after* a successful MQTT publish no longer regresses the row to `failed` — the data is already on the cloud, so we only surface a soft info toast and log the local issue.

`sendMqttEvent` got a 15s hard timeout (`Promise.race`) so AWS Cognito / paho-mqtt hangs surface as normal failures instead of silent limbo.

## OJD-1515 — Stop saturating AWS Cognito

Three issues in `create-mqtt-connection.ts`:

1. `GetIdCommand` was called on **every** MQTT publish. Each call mints a fresh Cognito identity. Now persisted in AsyncStorage keyed by `cognito_identity_id:<IdentityPoolId>`; stale identities are cleared and re-minted on `ResourceNotFoundException` / `NotAuthorizedException`.
2. AWS credentials were re-fetched on every publish too. New in-memory cache respects SDK-reported `Expiration` minus a 60s safety margin; refresh-once-then-share promise lock for concurrent callers. **Credentials never persisted to disk** — only the public `IdentityId` is on disk.
3. Mobile batch uploader concurrency 10 → 3 to flatten the post-field burst.

A phone uploading 200 measurements in a session: was 200 `GetId` + 200 `GetCredentials` → now 0 + 1.

## OJD-1516 — `pending` status + CHECK constraint

`MeasurementStatus` enum is now `pending | uploading | failed | successful`. `pending` distinguishes never-attempted from genuinely-failed so field metrics stop conflating them.

`apps/mobile/drizzle/0001_add_pending_status.sql` — SQLite table-recreate (SQLite can't add CHECK in place). Existing rows keep their statuses (all still valid). The CHECK enforces the four values at the DB level.

Side improvements:
- `getMeasurements` accepts `MeasurementStatus | MeasurementStatus[]` and returns `{ id, status, data }` rows.
- `useAllMeasurements` now pushes the active filter into the SQL query (no JS-side filtering) and reads counts from a dedicated `COUNT GROUP BY` SQL query (`countMeasurementsByStatus`). No JS state duplicating DB data.
- UI surfaces the raw storage status — no lossy `synced`/`unsynced`/`syncing` mapping. `pending` rows get their own `CloudUpload` icon distinct from `failed`'s alert.
- `resetUploadingMeasurements` resets interrupted `uploading` rows to `pending` (they were never confirmed failed) instead of `failed`.
- List-cache query keys: `setQueryData(["measurements"], …)` was a no-op against the new `["measurements", "list", filter]` cache; replaced with `invalidateQueries({ queryKey: ["measurements"] })` which prefix-matches every measurements-derived cache.

## UX polish

Delete-confirmation copy now distinguishes the two outcomes — they were both reading as "delete from local storage" but the consequences are very different:
- **Synced** → *"Remove from phone"* — explains the cloud copy stays intact.
- **Unsynced** → *"Delete unsynced measurement"* — warns the measurement will be lost permanently.

## CI gate for typecheck

A latent typecheck regression on this branch was caught only after running `tsc` locally — CI ran lint + test but never type-checked the mobile app. To prevent recurrence:

- Added a **Type Check Packages** step to `pr.yml` that runs `pnpm turbo run check-types` for the affected apps.
- `apps/mobile/package.json` now exposes a `check-types` alias (`tsc --noEmit`) so the existing turbo task picks it up — matches `apps/web`'s convention.
- Fixed three latent type errors that this exposed on `main`:
  - `FlowNodeType` was missing `"branch"` (the API schema already emits it).
  - `uniq()` used `[...new Set(...)]` which needs `target >= es2015`; switched to `Array.from(...)`.
  - Mobile tsconfig now disables `declaration`/`declarationMap` (mobile is an app, not a library — the `@ts-rest` contract type defeats declaration emit with TS2742 + TS7056).

## Test plan

- [x] `pnpm vitest run` — **362/362 passing**.
- [x] `pnpm lint` — clean.
- [x] `pnpm turbo run check-types --filter=mobile` — clean.
- [x] Migration upgrade-path test: seed v0 schema, apply `0001`, verify (a) rows preserved with original statuses, (b) non-status columns preserved verbatim, (c) CHECK activates post-migration, (d) `pending` accepted, (e) safely re-runnable.
- [x] Cognito caches: 9 cases covering GetId persistence, AsyncStorage hit, stale-identity recovery, concurrent first-time callers sharing one fetch, credentials reuse, expiration refresh (fake timers), safety-margin refresh, concurrent-fetch dedup, eviction-on-failure.
- [x] Regression test: `markUploaded` failure after successful publish does NOT flip row to `failed` (covered in `use-measurements`, `use-measurement-upload`, `use-questions-upload` test suites).
- [x] Manual: install on device, take measurements with airplane mode on → confirm they appear in Recent instantly with `pending` icon → toggle airplane mode off → confirm they flip to `successful`.
- [ ] Manual: confirm flag UI is reachable on `pending` rows.
- [ ] After deploy: monitor Cognito Identity Pool size to confirm no new junk identities are minted per publish.

## Notes / follow-ups (not in this PR)

- **Flag-after-sync** — once flagged on the device fails (currently you can't flag a `successful` row). Petar mentioned a ticket may already be in flight; out of scope here.
- **New filtering / visualization view** — exists per Petar; not wired up in this PR.
- The four-value enum opens the door to surfacing pending count separately in the toolbar if useful for field operations.

### DIFF
--- a/apps/mobile/src/hooks/__tests__/use-measurement-upload.test.ts
+++ b/apps/mobile/src/hooks/__tests__/use-measurement-upload.test.ts
@@ -10,6 +10,7 @@ const {
   mockExportSingle,
   mockToastSuccess,
   mockToastError,
+  mockToastInfo,
   mockShowAlert,
 } = vi.hoisted(() => ({
   mockSaveMeasurement: vi.fn(),
@@ -19,6 +20,7 @@ const {
   mockExportSingle: vi.fn(),
   mockToastSuccess: vi.fn(),
   mockToastError: vi.fn(),
+  mockToastInfo: vi.fn(),
   mockShowAlert: vi.fn(),
 }));
 
@@ -54,7 +56,7 @@ vi.mock("~/utils/measurement-annotations", () => ({
 }));
 
 vi.mock("sonner-native", () => ({
-  toast: { success: mockToastSuccess, error: mockToastError },
+  toast: { success: mockToastSuccess, error: mockToastError, info: mockToastInfo },
 }));
 
 vi.mock("~/components/AlertDialog", () => ({
@@ -209,6 +211,30 @@ describe("useMeasurementUpload", () => {
 
     expect(mockSendMqttEvent).not.toHaveBeenCalled();
     expect(mockSaveMeasurement).not.toHaveBeenCalled();
+  });
+
+  // Regression: a markUploaded failure that happens AFTER a successful MQTT
+  // publish must not flip the row back to "failed" — the data is on the
+  // cloud, only the local status write failed. The user sees an info toast,
+  // not the "upload not available" error toast.
+  it("does not mark the row failed when markUploaded errors after a successful publish", async () => {
+    mockSaveMeasurement.mockResolvedValueOnce("row-1");
+    mockSendMqttEvent.mockResolvedValueOnce(undefined);
+    mockMarkUploaded.mockRejectedValueOnce(new Error("disk full"));
+    const consoleSpy = vi.spyOn(console, "error").mockImplementation(vi.fn());
+
+    await capturedCallback(baseArgs);
+
+    expect(mockMarkFailed).not.toHaveBeenCalled();
+    expect(mockToastError).not.toHaveBeenCalled();
+    expect(mockToastSuccess).not.toHaveBeenCalled();
+    expect(mockToastInfo).toHaveBeenCalledWith("Uploaded — local status will refresh on next sync");
+    expect(consoleSpy).toHaveBeenCalledWith(
+      "Local status update failed after successful publish:",
+      expect.any(Error),
+    );
+
+    consoleSpy.mockRestore();
   });
 
   it("logs the upload error", async () => {
--- a/apps/mobile/src/hooks/__tests__/use-measurement-upload.test.ts
+++ b/apps/mobile/src/hooks/__tests__/use-measurement-upload.test.ts
@@ -4,6 +4,8 @@ import { useMeasurementUpload } from "../use-measurement-upload";
 
 const {
   mockSaveMeasurement,
+  mockMarkUploaded,
+  mockMarkFailed,
   mockSendMqttEvent,
   mockExportSingle,
   mockToastSuccess,
@@ -11,6 +13,8 @@ const {
   mockShowAlert,
 } = vi.hoisted(() => ({
   mockSaveMeasurement: vi.fn(),
+  mockMarkUploaded: vi.fn(),
+  mockMarkFailed: vi.fn(),
   mockSendMqttEvent: vi.fn(),
   mockExportSingle: vi.fn(),
   mockToastSuccess: vi.fn(),
@@ -21,7 +25,11 @@ const {
 let capturedCallback: (...args: any[]) => Promise<any>;
 
 vi.mock("~/hooks/use-measurements", () => ({
-  useMeasurements: () => ({ saveMeasurement: mockSaveMeasurement }),
+  useMeasurements: () => ({
+    saveMeasurement: mockSaveMeasurement,
+    markUploaded: mockMarkUploaded,
+    markFailed: mockMarkFailed,
+  }),
 }));
 
 vi.mock("~/services/mqtt/send-mqtt-event", () => ({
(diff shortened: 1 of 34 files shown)


apps/mobile/src/hooks/__tests__/use-measurement-upload.test.ts
```diff
@@ -198,10 +213,34 @@ describe("useMeasurementUpload", () => {
     expect(mockSaveMeasurement).not.toHaveBeenCalled();
   });
 
+  // Regression: a markUploaded failure that happens AFTER a successful MQTT
+  // publish must not flip the row back to "failed" — the data is on the
+  // cloud, only the local status write failed. The user sees an info toast,
+  // not the "upload not available" error toast.
+  it("does not mark the row failed when markUploaded errors after a successful publish", async () => {
```
xiduzo: i dont get this test? after a successfull upload it should never be able to reach a path that can mark it as failed?!

apps/mobile/src/hooks/__tests__/use-measurements.test.ts
```diff
@@ -136,41 +155,67 @@ describe("useMeasurements", () => {
...
+
+      consoleSpy.mockRestore();
+    });
+
+    // Regression: the previous implementation called
+    // setQueryData(["measurements"], ...), which does NOT match the list cache
+    // keyed ["measurements", "list", filter] read by the Recent screen, so
+    // in-place status updates silently missed visible rows. The fix relies on
+    // invalidateQueries({ queryKey: ["measurements"] }) which prefix-matches
+    // every measurements-derived cache.
```
xiduzo: i dont care what the previous implementation did - I care about what it does (/ should do) now

apps/mobile/src/hooks/__tests__/use-questions-upload.test.ts
```diff
@@ -127,12 +146,35 @@ describe("useQuestionsUpload", () => {
...
 
     consoleSpy.mockRestore();
   });
 
-  it("shows error toast when local storage also fails", async () => {
-    mockSendMqttEvent.mockRejectedValueOnce(new Error("offline"));
+  // Regression: a markUploaded failure that happens AFTER a successful MQTT
+  // publish must not flip the row back to "failed" — the answers are on the
+  // cloud, only the local status write failed.
+  it("does not mark the row failed when markUploaded errors after a successful publish", async () => {
```
xiduzo: this test is double?

apps/mobile/src/services/mqtt/create-mqtt-connection.ts
```diff
@@ -89,36 +90,184 @@ export async function createSignedUrl(params: {
...
-async function getCredentials({
-  region,
-  identityPoolId,
-}: {
-  region: string;
-  identityPoolId: string;
-}) {
-  const identityClient = new CognitoIdentityClient({ region });
+// Cognito IdentityId is durable per device + pool — once issued it can be
+// reused forever. Calling GetId on every connection creates a brand-new
```
xiduzo, line 94: Is there no invalidation on identity? e.g. should we refetch on app forgrounded or something similar?

apps/mobile/src/services/mqtt/send-mqtt-event.ts
```diff
@@ -1,6 +1,15 @@
 import { createMqttConnection, ReceivedMessage } from "~/services/mqtt/create-mqtt-connection";
 
-export async function sendMqttEvent(topic: string, payload: object) {
+// paho-mqtt + the Cognito handshake can hang silently when the device is
+// offline or AWS is throttling. Without a hard ceiling the caller's promise
+// never settles, leaving the measurement in limbo (not saved, not failed).
```
xiduzo: this is not true anymore? we now save the measurement before we try and sent it over MQTT?

apps/mobile/src/services/mqtt/create-mqtt-connection.ts
dominikvrbic: Nope, it should be one ID per device forever

vladstoe (approved):

PR #1467, opened 2026-05-14
feat(web): wire data-filter chip bar into the experiment data table with URL-persisted filters

petar-INFO:

- [x] Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Test Update
- [ ] Chore
- [ ] Other (please describe)

## Description

Wires the existing `FilterChipBar` into the experiment data table and persists filter state in the URL, so deep links and back/forward navigation reproduce the filtered view.

The table now reads filters from `useUrlDataFilters`, forwards them through `useExperimentData`, and snaps pagination back to page 0 whenever the active filter set changes (otherwise the UI shows "page 5 of 1" since the filtered backend path collapses to a single page).

`useExperimentData` gains a `filters` parameter. When non-empty, it JSON-encodes them into the request, drops `page`/`pageSize` (the ad-hoc backend path ignores them), and includes the serialized filters in the React Query key so distinct filter sets cache independently. Empty/draft filters are stripped via `zDataFilter.safeParse` before being sent or written to the URL.

Column ordering logic that previously lived inside `createTableColumns` is extracted to an exported `sortColumnsForDisplay` and also applied to `rawColumns` in the hook result. This keeps downstream consumers (filter picker, dashboard table-widget column picker) on the same "timestamps first, then by type precedence" order the data tab renders, instead of whatever raw order Databricks returns.

Two new hooks:

- `useUrlState`: generic URL-synced state with pluggable `serialize`/`parse`, `replace`/`push` method, and an opt-out for bidirectional sync. Preserves unrelated query params on write and skips the initial write when local state already matches the URL.
- `useUrlDataFilters`: thin wrapper that namespaces the param per table (`f_<tableName>`) and exposes a `completeFilters` view that excludes draft rows.

### DIFF
--- a/apps/web/components/experiment-data/experiment-data-table.tsx
+++ b/apps/web/components/experiment-data/experiment-data-table.tsx
@@ -20,6 +20,7 @@ import {
   formatValue,
   LoadingRows,
 } from "~/components/experiment-data/experiment-data-utils";
+import { useUrlDataFilters } from "~/hooks/useUrlDataFilters";
 
 import type { AnnotationType } from "@repo/api/schemas/experiment.schema";
 import { useTranslation } from "@repo/i18n";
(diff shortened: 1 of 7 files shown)


dominikvrbic (approved):
LGTM

PR #1469, opened 2026-05-14
feat: add per-experiment contributor anonymization (toggle, read path, export override)

petar-INFO:

- [x] Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Test Update
- [ ] Chore
- [ ] Other (please describe)

## Description

Adds a per-experiment contributor anonymisation setting. Owners flip the toggle on the experiment's visibility card; from then on every read path (data table, charts) and exports return stable `Contributor-XXXXXX` labels in place of real `id` / `name` and a nulled `avatar`.

### Anonymisation policy

`ContributorAnonymizerService` is the single seat for the rule on the API side. Pseudonyms are deterministic per `(experimentId, userId)` via the first 6 upper-hex chars of `sha256("{experimentId}:{userId}")`. Scoping the hash by experiment id prevents an owner of two experiments from correlating the same `Contributor-A4F2E1` across them and recovering identity.

The matcher keys off the existing `WellKnownColumnTypes.CONTRIBUTOR` literal, so generic `STRUCT<...>` columns pass through untouched. Rows are not mutated in place; a fresh array is returned.

`ExperimentDataRepository.transformSchemaData` now routes every page (paginated, fetch-all, macro) through the anonymiser, so charts and the data table inherit the policy without per-callsite work.

### Export path

The Databricks PySpark task (`apps/data/src/tasks/data_export_task.py`) mirrors the same hash in Spark SQL (`upper(substring(sha2(concat(experimentId, ':', col.id), 256), 1, 6))`) so exported CSV / NDJSON / JSON-array / parquet files agree with what the API would return. Matching is by full struct signature `(id, name, avatar): STRING` to stay consistent with the API's well-known-type check.

`InitiateExportBody` gains an optional `anonymizeContributors` override:
- Omitted: falls back to the experiment's stored setting.
- Present (`true` / `false`): wins over the stored value for that one export.

The resolved boolean flows through the use case, repository, and `DatabricksPort.triggerDataExportJob` down to the job widget (stringified `"true"` / `"false"` on the wire because Databricks widgets are string-only).

### Schema & contract

- Migration `0029_massive_lifeguard.sql` adds `experiments.anonymize_contributors boolean NOT NULL DEFAULT false`.
- `zExperiment` exposes the field; `zUpdateExperimentBody` and `zInitiateExportBody` accept it as optional.
- `ExperimentRepository` selects the column on both `findOne` and list reads.

### UI

- New `ExperimentAnonymizeToggle` on the visibility card writes the stored default via `useExperimentUpdate`.
- Export modal shows a checkbox seeded from the experiment's stored value; flipping it sets a per-export override and surfaces an "overrides experiment default" hint when the user diverges. The override is cleared when the modal closes.
- New en-US / de-DE / nl-NL strings under `experimentData.exportModal` and `experimentAnonymize`.


dominikvrbic (approved):

dominikvrbic (approved):
Lgtm

PR #1471, opened 2026-05-15
feat(web,viz): per-participant bar chart for classroom experiments

dominikvrbic:
## Summary

Adds a **Bar chart type** to the experiment visualisation workbench so teachers can build per-participant views for classroom measurements (count histograms, value averages, leaderboards) and group by question answers.

- Accepts categorical X columns and — uniquely — the well-known `CONTRIBUTOR` struct and `QUESTIONS` array. When X is a QUESTIONS array the data panel surfaces a secondary "Question" picker populated from the labels actually present in the data, and the renderer extracts each row's matching `question_answer` as the group key.
- Aggregation happens client-side: `count`, `sum`, `avg`, `min`, `max`. Null contributors / unanswered questions collapse into a single "Unknown" bucket.
- Vertical or horizontal orientation, sort by value, top-N — together those configure a "leaderboard" view.
- Existing line and scatter charts unchanged.

Also wires three seeded experiments to specific Databricks experiment IDs so `pnpm db:seed` produces a local DB pre-linked to real measurement data, with four placeholder contributor users keyed to the contributor UUIDs in those rows.

## Walkthrough video

A 1:34 / 1920×1080 walkthrough is posted as a PR comment so GitHub's inline video player works correctly. Six real-data scenarios across the Soil Health + Corn Photosynthesis experiments.

Direct download + fallback page: https://github.com/Jan-IngenHousz-Institute/open-jii/releases/tag/viz-bar-demo-2026-05-15

## Files of note

- `apps/web/components/experiment-visualizations/charts/bar/` — new bar chart definition (renderer, data panel, style panel, defaults, aggregation helpers + tests)
- `apps/web/components/experiment-visualizations/charts/registry.ts` — registers `bar`
- `apps/web/components/experiment-visualizations/charts/form-values.ts` — extends `ChartFormConfig` with bar-specific options (orientation, sort, topN, aggregation function, xColumnType, questionLabel)
- `apps/web/components/experiment-visualizations/workspace/visualization-workspace.tsx` — lets `CONTRIBUTOR` and `QUESTIONS` columns through the workspace's plottable filter
- `packages/api/src/utils/visualization-contracts.ts` — adds the `bar` role contract
- `packages/i18n/locales/{en-US,nl-NL,de-DE}/experimentVisualizations.json` — strings for bar options + the "Question" picker
[...]

PR #1475, opened 2026-05-16
chore(web,tests): fix accumulated check-types errors in test fixtures

dominikvrbic:
## Summary

`apps/web` had 94 accumulated `tsc --noEmit` errors across 41 test files where fixtures had drifted from the current contracts/types. This PR fixes them in one go — no production code touched, only test fixtures and a single factory function.

Categorised:

- **Session shape**: `createSession()` now returns the real `NonNullable<Session>` from `@repo/auth/types` (added required `user.createdAt`/`updatedAt`/`emailVerified`/`image` + the full session object; dropped invented `firstName`/`lastName` that were never on the session type).
- **Experiment status enum**: replaced `"completed"` / `"unknown"` with valid enum values; null/undefined experiment fixtures replaced with `createExperiment(...)`.
- **Contract status unions**: every `server.mount(... { status })` now uses a status the contract actually declares (e.g. 403→500 for delete/update endpoints, 400→404 for layout error paths).
- **Hook test bodies**: aligned with current contracts:
  - locations add/update wraps in `{ locations: [...] }`
  - metadata create/update drops invalid `location` / `key` fields
  - member role update returns the full member; role `"viewer"` → `"member"`
  - data upload responds with `{ uploadId, files }` (was `{ success }`)
  - location geocode body is now `Place[]` (was `{ place }`)
  - protocol compat macros + workbook version + viz update bodies use factory-built objects with all required fields
- **Component fixtures**: form steps add `step`/`goToStep`, review drops removed `protocols` field, navigation mobile uses `as const`, iot connection adds `bluetoothReason`/`serialReason`, branch cell adds `name`, protocol cells use `createProtocol()`, transfer-request uses `createTransferRequest()`.
- **Misc**: blog page test guards against undefined element, cookie settings casts the partial PostHog mock, breadcrumbs narrows the mocks union with `in`-guards, `instanceof`-on-primitive cases retyped as `unknown`, three chart renderer tests cast the strongly-typed `ChartFormConfig` to `Record<string, unknown>` at the factory seam.

## Test plan

- [x] `pnpm --filter web exec tsc --noEmit` — 0 errors (was 94)
- [x] `pnpm --filter web lint` — clean
- [x] `pnpm --filter web test` — vitest is green across all 41 touched files (no test logic changed, only fixtures aligned to current types)
- [x] `git diff --name-only main...HEAD | xargs -I{} grep -l "test" {} | wc -l` — only test files + `test/factories.ts` + `TESTING.md` are touched; no production code

## Why this is its own PR

Discovered while green-lighting the per-participant bar chart PR (#1471) for CI. The errors are all pre-existing on main — the bar PR inherited the same shape for one of its own test files but is otherwise unrelated. Splitting these out so the bar work and the broad-spectrum test-fixture refresh can review/merge independently.

## Out of scope

- Adding new test scenarios
- Touching production code
- Reformatting unrelated files (only modified files prettier-formatted)

### DIFF
--- a/apps/web/app/[locale]/(info)/blog/page.test.tsx
+++ b/apps/web/app/[locale]/(info)/blog/page.test.tsx
@@ -1,4 +1,4 @@
-import { render, screen } from "@/test/test-utils";
+import { assertExists, render, screen } from "@/test/test-utils";
 import { notFound } from "next/navigation";
 import { describe, it, expect, vi, beforeEach } from "vitest";
 import { getContentfulClients } from "~/lib/contentful";
@@ -71,6 +71,7 @@ describe("BlogLandingPage", () => {
 
   it("renders featured hero and post grid", async () => {
     const ui = await Page(params);
+    assertExists(ui, "Page should return a UI element for this fixture");
     render(ui);
     expect(screen.getByRole("region", { name: /featured hero/i })).toHaveTextContent(
       "Featured Post",
--- a/apps/web/test/factories.ts
+++ b/apps/web/test/factories.ts
@@ -35,6 +35,7 @@ import type {
 } from "@repo/api/schemas/workbook-cells.schema";
 import type { WorkbookVersionSummary } from "@repo/api/schemas/workbook-version.schema";
 import type { Workbook } from "@repo/api/schemas/workbook.schema";
+import type { Session } from "@repo/auth/types";
 
 // ── Experiment ──────────────────────────────────────────────────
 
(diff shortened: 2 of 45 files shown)


apps/web/app/[locale]/(info)/blog/page.test.tsx
```diff
@@ -71,6 +71,7 @@ describe("BlogLandingPage", () => {
 
   it("renders featured hero and post grid", async () => {
     const ui = await Page(params);
+    if (!ui) throw new Error("Expected Page to return a UI element");
```
Comment (petar-INFO) [inline apps/web/app/[locale]/(info)/blog/page.test.tsx]: .

apps/web/test/factories.ts
```diff
@@ -97,28 +98,37 @@ export function createExperimentAccess(
 
 // ── Session / Auth ──────────────────────────────────────────────
 
+type SessionShape = NonNullable<Session>;
```
petar-INFO: .

petar-INFO (approved):

PR #1503, opened 2026-05-20
fix(mobile): trim happy-path toasts; inline UI for QR mismatch

dominikvrbic:
Closes [OJD-1545](https://linear.app/openjii/issue/OJD-1545/mobile-trim-happy-path-toasts-replace-qr-mismatch-toast-with-inline-ui).

## Summary

- Removes 15 toasts that fired on the happy path. Each had a UI affordance already conveying the same state (row status tag, toolbar counts, calibration progress screen), so the toast was visual noise that crowded out the cases where a toast actually signals something.
- Drops both time-sync `warning` toasts (too noisy in the field). Failures still surface via the upload-error path.
- Replaces the QR-mismatch toast with an inline dismissible banner under the multi-choice search bar (\`accessibilityRole=\"alert\"\`, polite live region). The toast disappeared before users could react; the banner sticks until dismissed or a successful scan replaces it.

## What's removed

| Toast | Why |
|---|---|
| \`savedOffline\` (×2) | Row appears with a \"pending\" tag |
| \`answersUploaded\` / \`measurementUploaded\` | Row's status tag flips to \"synced\" |
| \`uploadedLocalStatusRefresh\` (×2) | Meta-toast about a local refresh users don't care about |
| \`uploadingMeasurements\` + \`measurementsSynced\` | Background auto-upload; user didn't trigger it, tags convey state |
| \`qrSelected\` + \`qrApplied\` (×2) | The scanned answer is already visible in the field |
| Calibration \`toastStarted\` + \`toastGainComplete\` + \`toastPanelMeasured\` | Calibration screen has its own progress UI |
| \`uploadAllSuccess\` | Toolbar counts + per-row tags already reflect success |
| \`uploadOneFailed\` (info) | Row flips to \"failed\" tag; toast was redundant |
| Time-sync \`toast.warning\` (×2) | Too noisy in the field; upload-error path still toasts |

## What's kept

Every error toast: \`uploadNotAvailable\`, \`uploadFailed\`, \`answersSaveFailed\`, \`saveToFileError\`, \`exportError\`, \`notConnected\`, \`noProtocol\`, \`scanError\`, \`errorConnect\`, \`errorDisconnect\`, \`toastProcessingError\`, time-sync \`unavailable\`, the global query-error fallback in \`configured-query-client-provider\`, and OTA update toasts.

## i18n

Removed the now-unused keys from \`en-US\` and \`nl-NL\` (\`recent-measurements.json\`, \`calibration.json\`, \`measurement-flow.json\`). Added \`questionNode.qrDismissLabel\` for the new banner's dismiss button.

## Tests

- Flipped the relevant \`expect(toast).toHaveBeenCalledWith(...)\` assertions to \`expect(toast).not.toHaveBeenCalled()\` and renamed test descriptions.
- 413 / 413 mobile tests pass, lint + typecheck clean.

## Test plan

- [ ] Record a measurement online → row appears as \"synced\" without a toast.
- [ ] Record a measurement offline → row appears with a \"pending\" tag, no toast.
- [ ] Trigger sync-all from the toolbar → counts update silently; only failures toast.
- [ ] Scan an invalid QR for a multi-choice question → inline red banner under the search bar shows the bad value; dismiss via X or re-scan.
- [ ] Scan a valid QR → answer applied silently.
- [ ] Calibration flow → screen progresses without toast clutter.

### DIFF
--- a/apps/mobile/src/features/calibration/screens/calibration/components/CalibrationFlow.tsx
+++ b/apps/mobile/src/features/calibration/screens/calibration/components/CalibrationFlow.tsx
@@ -42,7 +42,6 @@ export function CalibrationFlow({ protocol }: CalibrationFlowProps) {
     setMeasurements([]);
     setCurrentMeasurementIndex(0);
     setProcessedOutput(null);
-    toast.info(t("calibration:flow.toastStarted"));
   };
 
   const handleGainCalibration = async () => {
@@ -62,7 +61,6 @@ export function CalibrationFlow({ protocol }: CalibrationFlowProps) {
 
     console.log("Gain calibration completed");
     setCurrentStep("measurements");
-    toast.success(t("calibration:flow.toastGainComplete"));
   };
 
   const handleMeasurement = async (panelNumber: number) => {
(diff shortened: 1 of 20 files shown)


vladstoe (approved):
lgtm

PR #1510, opened 2026-05-21
feat(data,infra): ingest S3 large IoT payloads into Databricks pipeline

Blagoj95:

- [x] Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Test Update
- [ ] Chore
- [ ] Other (please describe)

## Description
Adds S3-based large IoT payload ingestion (>128 KB, uploaded via backend-issued pre-signed URL) to the Databricks medallion pipeline via Auto Loader directory listing. A new Bronze streaming table `raw_large_data` feeds the Silver layer through a dedicated DLT append flow into `clean_data`, so the existing two-source streaming checkpoint stays valid and no full refresh is needed on deploy (dev or prod). The earlier SQS/DLQ notification stack and its Grafana alerts are removed.

Security notes: the server-issued S3 key (`large-iot/{experimentId}/{uuid}.json`) is authoritative for `experiment_id` over the client-controlled payload field, and `client_id` stays null (pre-signed uploads never touch the broker, so the self-reported device_id is not promoted to a trusted identity).

## Related Issues

- Closes #OJD-1461
- Depends on #1303 

## Testing Plan

**Already validated on dev (2026-07-23):**
- `apps/data` unit tests: 64/64 pass (uv, JDK 17).
- Branch bundle deployed to the dev workspace; `raw_large_data` notebook registered on Centrum-DLT-Pipeline-DEV.
- Two synthetic payloads (~2.4 KB gzip'd JSON, one with `experiment_id` in the body, one relying only on the S3 key) dropped in `s3://open-jii-large-iot-dev/large-iot/00000000-0000-0000-0000-000000000dev/`.
- Pipeline update COMPLETED without a full refresh - the pre-existing `clean_data` checkpoint survived, confirming the append-flow structure works against live state.
- Rows verified end-to-end: 2 in `raw_large_data` (experiment_id correct in both, incl. the key-extraction case), 2 in `clean_data` (sample decompressed 2.4 KB -> 47 KB, GPS lat/lon intact, questions parsed, `client_id` null), 2 in gold `experiment_raw_data`.
- Negative path exercised implicitly: the pre-append-flow revision failed on dev with `STREAM_FAILED: 2 sources in checkpoint, 3 requested`, which is exactly what this structure prevents in prod.

**At merge (deploy verification):**
- [ ] Tofu apply (dev + prod): expect removal of the SQS queues/DLQ/S3-notification + Grafana large-IoT alert group, and the two stale pipeline conf keys (`LARGE_IOT_SQS_QUEUE_URL`, `LARGE_IOT_INGEST_ENABLED`); `databricks_large_iot_read` IAM policy must show NO destroy/create (state address stays `[0]`).
- [ ] Prod pipeline: first update after deploy completes without full refresh and the `clean_data_large_iot` flow appears in the DLT graph.
- [ ] Smoke test on prod (optional until a producer exists): drop one JSON under `large-iot/<real-experiment-id>/<uuid>.json` in the prod large-iot bucket, confirm the row reaches `clean_data` with the key-derived experiment_id.

**Out of scope here:** nothing produces large payloads yet - mobile wiring to `/api/v1/iot/upload-url` is OJD-1684.


Blagoj95:
@coderabbitai review

dominikvrbic (approved):

PR #1521, opened 2026-05-28
fix(mobile): tab badge count, persisted flow stores, slim flow screen

dominikvrbic:
## Summary

Consolidated mobile perf + flow-runner refactor (PR #1523 was folded in to keep this to a single review). All changes target the budget-phone "any button takes ~40s with ~20 measurements" symptom and the related architectural smells around the measurement-flow screen.

### Perf — tab badge

- `RecentTabIcon` rendered the unsynced badge by calling `useFailedUploads`, which read the full pending-or-failed list via `getMeasurements(["pending", "failed"])`. That function `pako.ungzip`s + `JSON.parse`s every row's compressed sample blob. Every measurement-cache invalidation kicked off N decompressions on the JS thread.
- Fix: `RecentTabIcon` now reads `useMeasurementCounts()` → `unsyncedCount` (single SQL `GROUP BY`, no decompression). Removed `useFailedUploads` and the `applySettledPatchBatch` branch that patched the now-unused `pending-or-failed` cache.

### Refactor — measurement-flow screen

- The screen pulled 9 fields out of `useMeasurementFlowStore` with a plain object destructure, so it re-rendered on every `scanResult` / `currentFlowStep` / `iterationCount` change and cascaded into `MeasurementFlowContainer` and its subtree. Now subscribes only to `experimentId`.
- The auto-pause and back-handler effects were extracted into `useAutoPauseFlow` / `useFlowBackHandler` hooks. Both use `useFocusEffect` so the cleanup IS the blur handler — no `isFocused` parameter, no early-return guard, zero `useEffect` in the screen body.
- `useExperiment(experimentId)` replaces `useExperiments()` + `.find(...)` so the screen subscribes to one cache key instead of running the full list query + per-item map on every render.

### Refactor — persist flow stores, drop usePausedFlowStore

The in-memory active + persisted snapshot split was complexity for its own sake when the active state is small (and historical data lives in SQLite). Collapsed:

- `useMeasurementFlowStore` now uses `persist` + AsyncStorage with a state-only `partialize`. Added `experimentLabel` as a first-class field set by the picker via `setExperimentId(id, label)`.
- `useFlowAnswersStore` is persisted the same way so answers survive a kill/background.
- Deleted `useAutoPauseFlow`, `usePausedFlowStore`, `PausedFlowSnapshot`, and the matching test. The pause action just navigates away (the state is already saved); resume just navigates back.
- `ExitFlowSheet.Pause` collapsed to `close() + dismissFlow()`. `Discard` calls `resetFlow + clearHistory` which write through `persist` and clear AsyncStorage.
- `useHomeContinueAction` collapsed to `router.push("/measurement-flow")`.
- `HomeContinueCard` subscribes to the flow store via `useShallow` and renders when `experimentId` is set and the flow has progressed past step 0 (or is on the questions-submit screen).
- `reset()` now routes through `resetFlow()` via `get()` so the persisted slice is fully cleared (cosmetic — `reset()` was only called from a unit test, but the contract is now defensible).
- `HomeContinueCard` falls back to `t("continue.untitled")` ("Measurement in progress" / "Meting bezig") when `experimentLabel` is missing — defensive for users with persisted state from before `experimentLabel` was a first-class field.

### Fix — bottom Android nav bar overlay

`AndroidNavigationBarBackground` painted a min-24px tinted bar at the bottom of every screen sized to `insets.bottom`. On devices with edge-to-edge gesture nav and a different inset (or `insets.bottom === 0`), it overlaid action-bar content (most visibly the analysis Accept Data button). Removed — system nav-bar coloring belongs on Android theming / `expo-navigation-bar`, not a JS overlay over content.

### UX shift to flag

With the persisted active flow store, tapping the Start Flow tab while an in-progress flow exists now lands on the in-progress flow (the persisted state IS the in-progress flow). To start fresh you need Discard via the Exit sheet. The Home Resume card is still the explicit "continue" entry.

## Test plan

- [ ] With ~20 stored measurements, tap around the app — every button responds within a frame, online or offline.
- [ ] Start a flow, advance steps, kill the app. Reopen → Home Continue card shows the experiment label + step count + Continue button. Tapping Continue lands you on the same step with the same scan result + answers.
- [ ] Exit sheet → Discard fully clears state (Home Continue card disappears, Start Flow lands on picker).
- [ ] Exit sheet → Pause leaves the screen; Home shows Continue card.
- [ ] Analysis cell → Accept Data still advances the flow (no regression from OJD-1546 fix).
- [ ] Greeting still shows the user's first name (no regression from OJD-1546 / #1520).
- [ ] Bottom of every screen: no grey bar overlaying content.
- [ ] `pnpm --filter mobile test` — 473 tests pass. Typecheck + lint clean.


apps/mobile/src/features/recent-measurements/components/recent-tab-icon.tsx
```diff
@@ -1,24 +1,29 @@
...
 import React from "react";
 import { View, Text } from "react-native";
-import { useFailedUploads } from "~/features/recent-measurements/hooks/use-measurements";
+import { useMeasurementCounts } from "~/features/recent-measurements/hooks/use-all-measurements";
 import { cn } from "~/shared/utils/cn";
 
 export function RecentTabIcon({ color, size }: { color: string; size: number }) {
-  const uploads = useFailedUploads();
-  const count = uploads?.length ?? 0;
+  // SQL GROUP BY count — does NOT decompress measurement payloads. Reading
```
petar-INFO: tell agent to tone down with annotations if possible

petar-INFO (approved):

PR #1523, opened 2026-05-28
refactor(mobile): scope measurement-flow-screen subscription and extract hooks

dominikvrbic:
## Summary

The flow screen pulled 9 fields out of `useMeasurementFlowStore` with a plain object destructure. Zustand's default is identity selector, so the screen re-rendered on every change to *any* field — `scanResult`, `currentFlowStep`, `iterationCount`, …  That cascade hit `MeasurementFlowContainer` and its children every flow tick and is a real factor in the budget-phone lag on the click-experiment → click-Start path.

- **Selective subscription.** The screen now subscribes only to `experimentId`. The auto-pause effect and the back handler don't need React subscriptions — they read store state via `getState()` at fire time. Net: no re-renders on flow progression, no cascading work.
- **Hook extraction.** `useAutoPauseFlow(isFocused, experimentLabel)` and `useFlowBackHandler(isFocused, hasActiveFlow)` live next to the other measurement-flow hooks. Effect dep lists shrink from 10 items to 2 each.
- **Faster snapshot clone.** `structuredClone(answersHistory)` replaces `JSON.parse(JSON.stringify(answersHistory))`. Same detach-from-live-store guarantee, faster on V8/Hermes.
- **Targeted label lookup.** `useExperiment(experimentId)` replaces `useExperiments()` + `.find(...)`. We were running the whole list query + the per-item map/transform on every render just to pull one label. Switching to the single-experiment query subscribes to one cache key and skips the list-wide map.

## Test plan

- [ ] Open and progress through a flow — title stays correct.
- [ ] Background the app mid-flow, foreground it — paused snapshot was saved, resume offered.
- [ ] Android hardware back during a flow opens the Exit sheet.
- [ ] On the picker (no active flow), the X exits as before.
- [ ] `pnpm --filter mobile test` (480 tests pass).


petar-INFO (approved):

dominikvrbic:
Folded into #1521 (merged `refactor/measurement-flow-screen` into `perf/mobile-tab-badge-count`). Shipping all the mobile-perf + flow-store work in one PR.

PR #1552, opened 2026-06-04
fix(mobile): support CH9102F USB-UART bridge for USB-serial sensors

vladstoe:

- [ ] Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Test Update
- [ ] Chore
- [ ] Other (please describe)

## Description

Adds support for connecting USB-serial sensors that use a **CH9102F USB-UART bridge**
(VID `0x1A86` / PID `0x55D4`), e.g. the Ambit. These previously failed on Android with
"could not connect" because the native `usb-serial-for-android` probe table didn't
recognize the chip.

## Changes Made

- Register the CH9102F in the prober's probe table via the existing patch, mapped to
  `CdcAcmSerialDriver` (the CH9102F is a USB CDC-ACM device).
- Unify the result/raw `TabBar` in `measurement-result` to the underline variant so it
  matches the recent-measurements toolbar.
- Minor layout tweak to the measurement no-questions empty state.

## Notes

- Native change: requires a fresh Android build (not OTA).
- ⚠️ Not yet tested on hardware (no Ambit on hand). If it connects but returns garbled
  data, baud rate (hardcoded 115200) is the likely next step.
- The ticket refers to the chip as "CH9012F", but VID/PID `0x1A86`/`0x55D4` is the
  **CH9102F** after some research. So, "CH9012F" is most likely a typo.

### DIFF
--- a/apps/mobile/src/features/measurement-flow/components/measurement-result/measurement-result.tsx
+++ b/apps/mobile/src/features/measurement-flow/components/measurement-result/measurement-result.tsx
@@ -137,7 +137,7 @@ export function MeasurementResult({
       {messageGroups.length > 0 && <MacroMessages messages={messageGroups} />}
 
       {/* Tab bar */}
-      <TabBar tabs={tabs} activeTab={activeTab} onTabChange={setActiveTab} />
+      <TabBar variant="underline" tabs={tabs} activeTab={activeTab} onTabChange={setActiveTab} />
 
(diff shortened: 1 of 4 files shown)


petar-INFO (approved):
lgtm lets deploy n see on the internal track

PR #1559, opened 2026-06-08
feat(mobile): force-update version gate from CMS

vladstoe:
- [x] Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Test Update
- [ ] Chore
- [ ] Other (please describe)

## Description 

CMS-driven force-update gate: when the running app version is below a Contentful-configured minimum, the app shows a full-screen, non-dismissable update screen instead of the navigator. The decision compares the live native version against a cached/fresh minVersion, so it works offline and clears automatically once the user updates.

### CMS

Added a new Contentful-driven force update model in `@repo/cms`.

This includes:

* New `PageForceUpdate` GraphQL query and fragment
* Regenerated SDK and schema
* Exported `PageForceUpdateFieldsFragment`
* New `PageForceUpdate` content type as a single entry

The content type contains the following fields:

* `title`
* `body` rich text
* `updateCta` using `ComponentButton` with label and store URL
* `minVersion` in `MAJOR.MINOR.PATCH` format
* optional `effectiveAt` schedule
* `active` kill switch

All fields are required except `effectiveAt`.

### Mobile gate logic

Added the force update gate logic on mobile.

This includes:

* `compare-version.ts` — small semver comparator for `MAJOR.MINOR.PATCH`, ignoring `-pre` and `+build`
* `fetch-force-update.ts` — fetches the single gate entry from Contentful and returns `null` if credentials are missing or nothing is published
* `use-force-update-gate.ts` — React Query hook that decides whether the user is `checking`, `allowed`, or `gated`

The gate is fail-safe: unparseable versions never gate, so a bad CMS value cannot brick users.

The hook uses `networkMode: "offlineFirst"` so offline refetches pause instead of erroring, allowing the cached gate to survive cold starts. It also re-checks on launch and resume, uses a 5 minute `staleTime`, reads the live native version through `expo-application`, and never gates in `__DEV__`.

### Force update UI

Added `ForceUpdateGate` and `ForceUpdateScreen`.

When the user is gated, `ForceUpdateGate` replaces the full app tree so navigation and background API calls do not run. The screen includes:

* Hero section
* Current → required version pills, for example `v1.1.0 → v1.2.0+`
* Scrollable rich-text body
* Pinned `Update` CTA opening `updateCta.url`

### Startup wiring

Updated `_layout.tsx` so `ForceUpdateGate` is mounted above the navigator.

Background services were moved into a new `AllowedAppServices` wrapper inside the gate, including OTA check, PostHog, TimeSync, Outbox, and related startup services. This prevents them from running while a user is gated.

Splash hiding is now centralized. `shouldHideSplash` also waits for i18n and force-update readiness, so the native splash remains visible until the gate has decided whether to show the force update screen or continue into the app. `RootLayoutNav` now reports navigation readiness up instead of hiding the splash itself.

### `CtfRichText` fixes

Updated the non-inline rich text rendering path. The alert banner is unaffected.

Changes include:

* Ordered lists now render numbers instead of bullets
* Lists are indented
* Blockquote and HR are visible in dark mode
* Marks no longer hardcode color
* Hyperlinks render colored instead of underlined

### Tests

Unit tests for `compare-version`, `useForceUpdateGate`, and `ForceUpdateGate`.

### Docs

`apps/docs/.../005-extending-the-platform/001-cms/003-force-update-gate.md`

## Checklist

- [x] I have updated the documentation accordingly


dominikvrbic (approved):
works nice!

PR #1560, opened 2026-06-08
fix(infra): large IoT S3 credential validation permissions and alert no-value fixes

Blagoj95:

- [ ] Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Test Update
- [ ] Chore
- [ ] Other (please describe)

## Description
Adds `s3:PutObject` and `s3:DeleteObjec`t to the Databricks storage credential policy — required by Databricks to validate the external location on creation. Adds `FILL(m1, 0)` to all three large IoT SQS alert rules so empty queues return 0 instead of no data, preventing false-positive `[no value]` firings

## Related Issues

- Closes #
- Related to #

### DIFF
--- a/infrastructure/modules/iot-core/main.tf
+++ b/infrastructure/modules/iot-core/main.tf
@@ -225,7 +225,7 @@ resource "aws_iam_policy" "databricks_large_iot_read" {
     Statement = [
       {
         Effect   = "Allow"
-        Action   = ["s3:GetObject", "s3:ListBucket"]
+        Action   = ["s3:GetObject", "s3:PutObject", "s3:DeleteObject", "s3:ListBucket"]
         Resource = [var.large_iot_bucket_arn, "${var.large_iot_bucket_arn}/*"]
       },
       {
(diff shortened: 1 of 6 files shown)


infrastructure/modules/iot-core/main.tf
```diff
@@ -225,7 +225,7 @@ resource "aws_iam_policy" "databricks_large_iot_read" {
     Statement = [
       {
         Effect   = "Allow"
-        Action   = ["s3:GetObject", "s3:ListBucket"]
+        Action   = ["s3:GetObject", "s3:PutObject", "s3:DeleteObject", "s3:ListBucket"]
```
petar-INFO: This will be a read-only bucket / ext. location so I think this is technically inappropriate .. so I would maybe suggest using this https://registry.terraform.io/providers/databricks/databricks/latest/docs/resources/external_location#read

It should hopefully adjust the UC permission probe from write+delete to a read-only check, which GetObject & ListBucket will suffice

petar-INFO (approved):

PR #1563, opened 2026-06-08
fix(mobile): surface protocol-unavailable distinctly and make precache recoverable (OJD-1564)

dominikvrbic:
## Why

Closes [OJD-1564](https://linear.app/openjii/issue/OJD-1564/intermittent-scan-error-blocks-offline-measurement-flow). Some users intermittently saw a generic **"Scan error"** that blocked the offline measurement flow; the only known workaround was going back online.

Root cause: a measurement node's protocol is fetched offline-first via `useProtocol(content.protocolId)`. When that protocol isn't in cache and the device is offline, `protocol` is `undefined`, and `measurement-node` threw `"No protocol"` which was caught and surfaced as the **generic** `scanError` toast — masking the real cause from both the user and telemetry. Reconnecting refetches the protocol, which is exactly why "go online" worked.

The cache is already persisted to disk, so the real gap is **coverage**: the per-experiment precache (`usePrecachedExperimentData`) used `prefetchQuery`, which **swallows per-asset errors**, so a partially-failed precache reported *success* and — with `staleTime: Infinity` — never retried for the session, leaving the protocol uncached until the next login.

## What

- **`measurement-node.tsx`** — replace the `"No protocol"` throw-into-generic-toast with an explicit guard that shows a specific `protocolUnavailable` message and logs the `protocolId`; disable the **Start** button while the protocol query is still loading so users can't tap into a guaranteed failure. The `try/catch` now wraps only `executeScan`, so `scanError` is reserved for genuine scan failures.
- **i18n** — add `measurementNode.toast.protocolUnavailable` to `en-US` and `nl-NL`.
- **`use-precached-experiment-data.ts`** — switch `prefetchQuery` → `fetchQuery` + `Promise.allSettled`, throwing if any asset fails (so the precache reports `error` instead of false success); drop `staleTime: Infinity` to a finite 5-min window so an incomplete precache becomes eligible to refetch on reconnect (via the global `refetchOnReconnect`) and fill the missing protocol without re-login; add `meta: { suppressToast: true }` so the now-possible offline error doesn't blare a technical toast.

## Tests

- `measurement-node.test.tsx` — protocol-unavailable shows the specific toast (not scan error) and skips `executeScan`; Start disabled while loading; scan success advances; genuine scan error preserved; no-device hides Start. 
- `use-precached-experiment-data.test.ts` — all-succeed caches protocols; a failed fetch → `isError` (the key new behavior); recovery via `refetch` (reconnect path); dedup of repeated protocol ids.

All 15 new tests pass; the changed files typecheck and lint clean. (Two unrelated pre-existing `main` failures — `ctf-rich-text.tsx` typecheck, `fetch-active-alerts.ts` lint — are untouched here.)

## Reviewer note

`OfflineModeIndicator` keys off `precachedData` truthiness, so it now appears only once **all** assets are cached (precache success) rather than after a partially-failed run — more accurate, but a visible behavior change.

## Out of scope (deferred follow-up)

Re-running the login prefetch on app foreground/reconnect (not just at login), per the ticket's noted follow-up.

### DIFF
--- a/apps/mobile/src/features/experiments/hooks/__tests__/use-precached-experiment-data.test.ts
+++ b/apps/mobile/src/features/experiments/hooks/__tests__/use-precached-experiment-data.test.ts
@@ -1,0 +1,115 @@
+// @vitest-environment jsdom
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import { renderHook, waitFor } from "@testing-library/react";
+import React from "react";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+import { usePrecachedExperimentData } from "../use-precached-experiment-data";
+
+const { mockGetFlow, mockGetProtocol, mockGetMacro } = vi.hoisted(() => ({
+  mockGetFlow: vi.fn(),
+  mockGetProtocol: vi.fn(),
+  mockGetMacro: vi.fn(),
+}));
+
+vi.mock("~/shared/api/tsr", () => ({
+  tsr: {
+    experiments: { getFlow: { query: (...a: unknown[]) => mockGetFlow(...a) } },
+    protocols: { getProtocol: { query: (...a: unknown[]) => mockGetProtocol(...a) } },
+    macros: { getMacro: { query: (...a: unknown[]) => mockGetMacro(...a) } },
+  },
+}));
+
+function flowWith(protocolIds: string[], macroIds: string[] = []) {
+  return {
+    body: {
+      graph: {
+        nodes: [
+          ...protocolIds.map((id) => ({ type: "measurement", content: { protocolId: id } })),
+          ...macroIds.map((id) => ({ type: "analysis", content: { macroId: id } })),
+        ],
+      },
+    },
+  };
+}
+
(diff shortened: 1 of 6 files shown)


dominikvrbic:
Closing as redundant. PR #1556 ("make scan errors observable") landed on main first and already covers the user-facing half of this work: the specific `protocolUnavailable` toast + guard, the i18n keys (en-US/nl-NL), and bumping the scan-error log from `warn` to `error` for PostHog.

Note for whoever picks up OJD-1564 next: #1556 deliberately deferred the **root-cause fix** (its own description: "makes the cause observable and improves the failure UX *while we confirm the root cause*"). The actual gap remains — `usePrecachedExperimentData` uses `prefetchQuery` (swallows per-asset errors) + `staleTime: Infinity`, so a partially-failed precache reports success and never retries for the session, leaving a protocol uncached offline until next login. The fix explored here (`fetchQuery` + `Promise.allSettled` + throw + finite `staleTime` so it recovers on reconnect), plus disabling Start while the protocol loads, is the unique remaining value if we want to fully close OJD-1564 rather than just instrument it.

PR #1565, opened 2026-06-08
feat(workbooks): create-from-experiment + go-live UX & in-use delete gate

dominikvrbic:
## OJD-1548 — Workbook go-live UX

Makes the workbook lifecycle coherent before go-live: create a workbook directly from an experiment (no more "link an existing one" detour), name/rename it, duplicate it, see where it's used, and guard the destructive actions.

### What's new
- **Create from experiment (M1):** the experiment flow tab can now create a workbook (create → auto-attach → open), not just attach an existing one.
- **Naming (M2):** dashboard "Create Workbook" prompts for a name instead of `Untitled Workbook - {date}`; inline rename via an editable title on the editor page.
- **Duplicate (M3):** "Copy of …" from the dashboard row menu, including workbooks you don't own (clone-to-edit).
- **Dashboard signal (M4):** per-type cell summary + a "used by N experiments" indicator (shared `WorkbookCellSummary`), powered by a computed `experimentCount` (correlated subquery in `findAll`/`findById`).
- **Safer delete (M5):** delete confirm warns when the workbook is attached to experiments.
- **Reproducibility guard (M6):** confirm dialog + stronger copy when changing/upgrading a live experiment's workbook.
- **In-use delete behind a flag:** deleting a workbook attached to experiments unlinks them and loses their measurement flow, so that action is now gated behind a new `workbook-deletion` PostHog flag (default off), mirroring `experiment-deletion`. Unused workbooks stay freely deletable.

### Screenshots

**Create from an experiment — attach existing *or* create new (M1)**
create or attach

**…created, auto-attached, and opened inline (M1)**
linked workbook card

**Dashboard "used by" column + `workbook-deletion` flag hiding Delete for an in-use workbook (M4 + flag)**
dashboard and flag gate

**Delete warning for an in-use workbook (M5; shown when the flag is enabled)**
delete in-use warning

**Change-workbook reproducibility guard on a live experiment (M6)**
change workbook confirm

### Testing
- New `workbook-list.test.tsx` covers the delete gate: flag-off hides Delete for in-use, keeps it for unused, flag-on shows it for in-use (3 passing).
- `tsc --noEmit` clean.
- Verified end-to-end against a local stack (seeded DB, real email-OTP login) with Playwright: M1–M6 each confirmed in the UI and DB, and the flag gate confirmed live (in-use → Delete hidden, unused → Delete shown).

### Notes
- Requires the `workbook-deletion` boolean flag in PostHog (created; default off).
- New experiments-flow strings are en-US only; de-DE/nl-NL fall back to en-US (consistent with existing workbook-attach keys). Workbook namespace is fully translated in all three locales.


petar-INFO (approved):

apps/backend/src/workbooks/core/repositories/workbook.repository.ts
```diff
@@ -16,6 +16,14 @@ export interface WorkbookFilter {
   userId?: string;
 }
 
+// Number of experiments currently referencing a workbook. A correlated subquery
```
petar-INFO, line 19: this seems wacky

dominikvrbic: agree but works good, tried to make it less but this ended up being the best

PR #1677, opened 2026-06-19
feat(workbooks): searchable picker, in-place rename, and editor polish (OJD-588)

### LINKED ISSUE #1650: Specific protocol causes experiment to keep alerting about workbook upgrade
## Bug report

Steps:

* Create a new or change existing experiment
* to attach a new workbook that has a protocol cell linked to protocol 'BF simple_fluor_light_curve_with_recovery 1'
* The Design tab of the experiment indicates that an upgrade is available
* Upgrade to the new version of the workbook
* Result: Upgrade is acknowledged, but the Design tab indicates another upgrade. This goes on indefinite times

Expected result: The upgrade is offered only once

see e.g [https://dev.openjii.org/en-US/platform/experiments/a9594cfa-1c19-4dff-a496-05ec8592f7a1/design](<https://dev.openjii.org/en-US/platform/experiments/a9594cfa-1c19-4dff-a496-05ec8592f7a1/design>)

## Root cause

`is-workbook-upgradable` compared the live workbook cells / referenced protocol code against the published version snapshot with `JSON.stringify`, which is key-order sensitive. Live entities keep object keys in insertion order, while jsonb snapshots come back from Postgres with keys re-normalised. For a protocol whose two orderings differ (the reported one), the two strings never match — so the workbook reads as "upgradable" forever, even immediately after upgrading. Hence the infinite loop, and why it is protocol-specific.

## What this PR changes

**Backend**

* `is-workbook-upgradable`: order-insensitive comparison (`stableStringify`) for the cell-design and protocol-code drift checks — the actual fix for the infinite loop. Macro check left unchanged (plain string).
* `attach-workbook` / `upgrade-workbook-version`: reject pinning a workbook version that does not belong to the target workbook (returns not-found) so an experiment can never be left pinned to a broken/foreign version.

**Frontend — experiment Design tab**

* The workbook is now editable in place when you have edit rights; the Edit/View toggle was removed (read-only for everyone without edit rights, no toggle either).
* Editing auto-applies: every autosave automatically publishes/re-pins the experiment to the latest workbook version — no manual "Upgrade" step.
* Removed the "workbook version upgraded" toast.
* `useDetachWorkbook` now invalidates the workbook + workbook-versions caches so the upgrade indicator never shows stale state after detach / re-attach.

**Behaviour notes**

* Auto-apply mints a new workbook version per edit-burst; no-op saves reuse the latest version (so they do not spam versions). The manual "updates available" banner still appears for external drift (a referenced protocol/macro changing outside the editor).

**Tests**

* Backend: regression test for the key-order canonical comparison (fails on `main`, passes with the fix) + version-ownership guard tests.
* Frontend: auto-apply-on-save, no-toggle (owner edits in place), detach cache invalidation, and toast-removal tests.

dominikvrbic:
This PR began as the OJD-1626 upgrade-prompt fix and now also carries the workbook authoring UX improvements from the 2026-06-22 field-researcher sync (OJD-588), plus docs and a walkthrough video.

# Workbook authoring UX (OJD-588)

Hands-on feedback from a field researcher after building ~10 workbooks. The concrete, low-risk items:

- **Searchable workbook picker.** A new reusable `WorkbookSelect` (Popover + cmdk) replaces the plain `Select` both when creating an experiment and when changing the attached workbook. It filters by name and stays collision-safe for duplicate names. (The "I can fix this now" item from the call.)
- **Rename in place.** The workbook owner can rename a linked workbook straight from the experiment Design tab instead of going back to the Workbooks list. Owner-only, since the update API is owner-gated.
- **Clearer cell sidebar.** Each cell shows its own name with a per-type icon and colour instead of a repeated "Question / Question / Question".
- **Prominent question field.** The required "Question text" field now stands out from the cell's data-column name.
- **Option character limit explained.** The multiple-choice editor labels the bare `64` counter with its reason (keeps choices readable as buttons in the mobile app).

i18n keys were added to en-US; de-DE and nl-NL fall back to en-US, matching the existing partial locales.

### Docs and walkthrough video

Updated the Workbooks guide and the changelog in `apps/docs` with an embedded Playwright walkthrough recording and stills, and corrected the stale "View pinned / Edit toggle" wording to the current inline-edit and auto-re-pin behaviour (the OJD-1626 part below). The recording spec is at `apps/web/e2e/record_workbook_ux.py`; it depends on the `apps/web/e2e` helpers harness landing via `chore/web-e2e-and-agents-md`.

### Testing

Affected web suites are green (379 tests across the workbook, experiment-flow and new-experiment areas, including a new owner-only rename test); `tsc` and eslint are clean; `pnpm --filter docs build` passes (onBrokenLinks is set to throw). The walkthrough was recorded against this branch's running stack.

Deliberately deferred as design work (also raised on the call): the versioning lifecycle, shared-workbook templates / branching, workbook visibility scoping, and multi-sensor workbooks.

Ref: https://linear.app/openjii/issue/OJD-588

---

# Upgrade-prompt fix (OJD-1626)

## What & why

The experiment **Design** tab kept offering the same workbook upgrade forever (OJD-1626). `is-workbook-upgradable` compared the live workbook cells / referenced protocol code against the published **jsonb** snapshot with `JSON.stringify`, which is key-order sensitive — live objects keep insertion order, jsonb comes back from Postgres re-normalised, so a protocol whose two orderings differ always read as "upgradable", even immediately after upgrading. Hence the loop, and why it was protocol-specific.

## Changes

**Backend**
- `is-workbook-upgradable`: order-insensitive `stableStringify` comparison for the cell-design and protocol-code drift checks (the actual fix). Macro check unchanged (plain string).
- `attach-workbook` / `upgrade-workbook-version`: reject pinning a workbook version that does not belong to the target workbook (not-found) so an experiment can't be left pinned to a broken/foreign version.

**Web — experiment Design tab**
- Edit the workbook in place when you have edit rights; removed the Edit/View toggle (read-only otherwise, no toggle either).
- Auto-apply: each autosave automatically re-pins the experiment to the latest workbook version — no manual "Upgrade" step.
- Removed the "workbook version upgraded" toast.
- `useDetachWorkbook` invalidates the workbook + workbook-versions caches so the upgrade indicator never goes stale after detach / re-attach.

## Behaviour notes
- Auto-apply mints a new workbook version per edit-burst; no-op saves reuse the latest version. The manual "updates available" banner still appears for external drift (a referenced protocol/macro changing outside the editor).

## Testing
- **Backend:** regression test for the key-order canonical comparison (fails on `main`, passes here) + version-ownership guard tests.
- **Web:** auto-apply-on-save, no-toggle (owner edits in place), detach cache invalidation, toast removal. Full web suite green (4422 tests); web line coverage ~92%.
- **Manual:** recorded the flow with Playwright (OTP login → Design tab → upgrade once → banner stays gone after reload), run against this branch's backend. Demo video + screenshots are on the Linear ticket.

Closes #1650
Ref: https://linear.app/openjii/issue/OJD-1626/specific-protocol-causes-experiment-to-keep-alerting-about-workbook


dominikvrbic:
### Demo recording

A Playwright recording of the fix (OTP login → experiment Design tab → accept the upgrade once → banner **stays gone after reload**), run against this branch's backend:

- **Video + before/after screenshots are attached to the Linear ticket:** [OJD-1626](https://linear.app/openjii/issue/OJD-1626/specific-protocol-causes-experiment-to-keep-alerting-about-workbook)
  - `Demo: upgrade offered once, then stays gone`
  - `Before: "Workbook has updates available" (on v1)` → `After: pinned to v2, no banner (persists after reload)`

> Note: GitHub only accepts inline video via the web composer (drag-and-drop), not via API/CLI, so I couldn't embed the `.webm` directly in this comment. To inline it here, drag the file into this box in the browser.

**Caveat on fidelity:** the exact reported bug is data-specific (a protocol whose stored jsonb key order differs from the live row), which doesn't exist in the local seed — so this demonstrates the *fixed* behavior (upgrade offered once, auto-applies, stays gone) rather than reproducing the original infinite loop. The root-cause regression is proven by the backend unit test (red on `main`, green here).

dominikvrbic:
### Updated demo (covers the actual changes)

Re-recorded against this branch's web+backend. Shows the **new** behavior, not the old manual flow:
- Workbook is **editable in place — no Edit/View toggle**.
- Each autosave **auto-pins** the experiment to a new version (v1 → v2 → v3) with **no manual "Upgrade" click and no toast**.
- After reload it stays on the latest version — **no perpetual "upgrade available" prompt**.

Video + screenshot attached on the Linear ticket: [OJD-1626](https://linear.app/openjii/issue/OJD-1626/specific-protocol-causes-experiment-to-keep-alerting-about-workbook) — *"Demo (new behavior): edit-in-place, no toggle, auto-apply on each save"* (supersedes the earlier offered-once recording). GitHub still doesn't allow embedding video via API, so it lives on the ticket.

vladstoe:
On auto-apply: editing the workbook only needs ownership, but the auto-upgrade after each save needs experiment-admin. So a non-admin owner can edit and the workbook saves fine, but the upgrade fails and they get an error toast on every save.

Should a non-admin owner be able to upgrade the workbook when editing it in the experiment "design" tab?

There's also a brief flicker: when auto-saving, the "updates available" banner flashes on for a split second before disappearing.

dominikvrbic:
only an admin should be able to edit or update a workbook related thing, and last one should be fixes as well

vladstoe (approved):

PR #1701, opened 2026-06-25
feat(backend, web): per-entity + global full-text search with ranking & fuzzy matching

vladstoe:

- [x] Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Test Update
- [ ] Chore
- [ ] Other (please describe)

### Why

Every search was a single-column `ILIKE '%term%'` on `name`. That meant:

- **No typo tolerance** — searching `experimnt` found nothing.
- **No stemming** — searching `running` didn't match an experiment called "Run".
- **No prefix matching** — `biolu` didn't find "Bioluminescence".
- **No relevance ranking** — when several names matched the same substring, results came back in default order (e.g. by `updatedAt`), not best-match-first.
- **No real global search** — the ⌘K palette only did hardcoded navigation, not search.

This PR fixes all of the above and adds a working global search across experiments, protocols, and macros, wired into the command palette.

**Approach:** PostgreSQL FTS (`tsvector`/`tsquery`) + `pg_trgm` fuzzy matching. No new infrastructure — just a migration plus Drizzle/NestJS code. Single fixed `'english'` config gives stemming + stop-word removal; the language of stored data is irrelevant.

### What you can now do

| Query | Matches | What's new |
|---|---|---|
| `chlorphyll` | "Chlorophyll fluorescence assay" | typo tolerance (trigram) |
| `flowering` | "Flower onset trial" | English stemming |
| `drought wheat` | "Wheat drought-stress study" | multi-word, any order |
| `nitrogen` | name matches first, then description matches, then experiments matched only via creator/member/location | relevance ranking |
| a collaborator's name | experiments they created or are a member of | cross-table matching |

### Database (migration 0036)

- Adds a generated `STORED` `search_vector tsvector` on `experiments`, `protocols`, `macros`, weighting **name (A) above description (B)**.
- GIN indexes on each vector; trigram GIN indexes (`gin_trgm_ops`) on entity names plus `profiles.first_name/last_name` and `users.email` for fuzzy matching.
- Enum fields (`protocols.family`, `macros.language`) are matched at query time, not stored in the vector — enum→text casts aren't `IMMUTABLE` and can't live in a generated column. The hand-edited SQL is documented inline.

### Backend

- **Shared FTS helper** (`fts.ts`): `buildTsQuery` (sanitizes input → safe prefix tsquery, e.g. `"foo bar!"` → `"foo:* & bar:*"`), `ftsMatch`/`ftsRank` (vector match + trigram typo tolerance + relevance score), `trigramMatch`, and `escapeLike` (escapes `%`/`_`/`\` so user input can't act as LIKE wildcards).
- **The experiment, protocol, and macro repositories** now search across name, description, creator name, and entity metadata — and, for experiments, member and location names via correlated subqueries. **The user repository** searches first name, last name, full name, and email (it has no description or creator).
- **Deactivated and soft-deleted accounts are excluded** from creator/member name matching everywhere (`activated = true AND deletedAt IS NULL`) — searching a deleted user's name won't surface their experiments.
- **New global search module** (`search/`): `GET /api/v1/search` runs the three entity searches in parallel, then **merges them by positional rank within each type** (each entity's per-type relevance score isn't comparable across types, so the top hit of each type ranks equally) and returns a ranked, merged list.

### Frontend

- **Command palette** is now a real search UI: controlled input, debounced query via `useGlobalSearch`, server-ranked results grouped into Experiments/Protocols/Macros with loading/empty/error states, selecting navigates to the entity. Static Pages/Actions are still filtered client-side.
- Supporting bits: `search-result-item.tsx`, a `commandProps` passthrough on the shared `command.tsx` (`shouldFilter={false}` since the server already ranks), contract/schema registration, and navigation i18n strings (en/de/nl).
- Per-entity list pages (experiments/protocols/macros) improve automatically — they already send `search` to the backend, so they inherit ranking + fuzzy matching with no component changes.
- Minor responsiveness fix to `SelectTrigger` / the experiments-list input width.

### Tests

Pure unit tests for query sanitization; real-Postgres repository + use-case specs covering prefix matching (`biolu` → "Bioluminescence"), typo tolerance (`experimnt`), stemming, name-beats-creator ranking, access control, and the deactivated/deleted-account exclusions; a field-coverage spec; controller and command-palette tests.


apps/backend/src/search/search-field-coverage.spec.ts
```diff
@@ -0,0 +1,348 @@
+import { assertSuccess } from "../common/utils/fp-utils";
```
petar-INFO: This test seems misplaced

Is it testing the entity repository changes (exp, macro etc.)?

dominikvrbic:
workbooks search still missing

vladstoe:
> workbooks search still missing

yep working on it

dominikvrbic (approved):

PR #1738, opened 2026-07-03
fix(viz): dashboard chart sizing, WebGL exhaustion, and correctness fixes

petar-INFO:

- [ ] Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Test Update
- [ ] Chore
- [ ] Other (please describe)

## Description

A batch of fixes for how charts render and stay correct on a dashboard, where they are small, dense, and driven by saved state rather than the live editor.

### Charts rescale to their container

Every non-cartesian and scientific wrapper (polar, radar, ternary, wind-rose, carpet, heatmap, correlation matrix, contour, density, ridge, parallel-coordinates/categories, alluvial, histogram-2d, SPC) now reads its container size and adapts to it; previously only cartesian charts did, so these rendered at fixed sizes and overflowed or clipped in dashboard cells. Two shared helpers back this: `responsiveChrome` (the coordinate-agnostic title/legend/margin/font slice of the base layout, so polar/ternary/carpet charts get tier-aware chrome without inheriting cartesian axes) and `tierAxisFontSizes` (per-tier tick and axis-title sizes). Long category tick labels are ellipsized per tier so Plotly's automargin can't grow until it eats the plot area, and past a tier's tick cap every Nth category is sampled. Faceted grids gain an ultra-compact tier that drops cell/axis/shared titles and keeps ticks only.

### WebGL context exhaustion

Grouped cartesian charts forced a WebGL trace per category, and a dashboard full of them exhausted the browser's WebGL context pool, making the data layer vanish on context loss. WebGL now auto-enables only past a point-count threshold, so small grouped charts stay on SVG and cost no context.

### Correlation matrix fetched wrong data from saved state

The corr aggregation was only seeded while editing, so a saved dashboard fetched raw rows and rendered an all-NaN matrix. The renderer now derives the pairwise corr functions itself, sharing one helper with the shelf so both agree on the projection.

### Contributor anonymization

Toggling anonymization re-pseudonymizes ids and names server-side. The update hook now invalidates the cached distinct-values and visualization-data reads so filters and charts refresh without a hard reload; a categorical filter drops a selected contributor id once the reloaded list no longer contains it (guarded to complete lists); and the color-map keys and labels overrides by contributor name so they keep matching what the chart looks up. STRUCT-to-name flattening is unified in one converter.

### Other correctness fixes

- Categorical color with multiple Y series: legendgroup keyed per (series × category) so every combo gets its own legend entry instead of collapsing to the first.
- Faceted dual-Y: secondary-axis series route onto a per-cell overlay axis instead of colliding with the primary grid axes.
- Numeric heatmap/contour and carpet axes sort ascending regardless of row order; shuffled rows previously scrambled iso-lines and non-monotonic carpet axes crashed Plotly's `contourcarpet`.
- Carpet `reverseScale` now reaches the trace and `showGrid` changes re-render (missing memo dependency).

### DIFF
--- a/apps/web/components/data-filters/use-distinct-options.ts
+++ b/apps/web/components/data-filters/use-distinct-options.ts
@@ -6,14 +6,14 @@ import { WellKnownColumnTypes } from "@repo/api/schemas/experiment.schema";
 import { useExperimentDistinctValues } from "../../hooks/experiment/useExperimentDistinctValues/useExperimentDistinctValues";
 
 export function useDistinctOptions(column: DataColumn, experimentId: string, tableName: string) {
-  const { values, truncated, isLoading, error } = useExperimentDistinctValues({
+  const { values, truncated, isLoading, isSuccess, error } = useExperimentDistinctValues({
     experimentId,
     tableName,
     column: column.name,
   });
   const isContributor = column.type_text === WellKnownColumnTypes.CONTRIBUTOR;
   const contributorMap = useContributorIdMap(values, isContributor);
-  return { values, truncated, isLoading, error, isContributor, contributorMap };
+  return { values, truncated, isLoading, isSuccess, error, isContributor, contributorMap };
 }
 
 export function useContributorIdMap(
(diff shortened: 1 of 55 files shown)


dominikvrbic (approved):
LGTM

PR #1751, opened 2026-07-07
feat(backend): Phase 1 — Better Auth organization layer + personal orgs

vladstoe:

- [x] Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Test Update
- [ ] Chore
- [ ] Other (please describe)

## Description

**Phase 1 of the unified IAM / organization-scoped access control feature (OJD-1638).**

This lands the entire Better Auth **organization plugin** (orgs + teams + invitations) and auto-provisions every user a **personal organization** they own — with **zero change to who can see or edit what**. All the custom authorization logic that actually gates access (per-resource grants, visibility, base permissions, list-scoping) is intentionally deferred to later phases; nothing here reads the new tables for access decisions.

It also **retires the legacy institution-org concept** (the free-text "organization" field from registration and its `profiles.organizationId` column) for a clean slate, since the new model fully replaces it.

**Why it's safe to ship:** new tables + nullable/defaulted columns only, the org plugin only *adds* routes, provisioning is additive and idempotent, and `activeOrganizationId` is written but never read for access. Existing experiment/macro/protocol/workbook access paths are untouched.

### Organization plugin & provisioning
- Enabled the Better Auth `organization` plugin (teams on) in `packages/auth/src/server.ts` and mapped the new tables through the drizzle adapter. Added `organizationClient` to both the web and native auth clients.
- New `packages/database/src/organizations.ts`: `ensurePersonalOrganization` (idempotent create-org-and-owner-membership, keyed by a deterministic `personal-<userId>` slug) + `personalOrgName`/`personalOrgSlug` helpers.
- Provisioning wired through three idempotent paths: `user.create.after` (signup), `session.create.before` (login self-heal + sets active org), and a one-off backfill script.
- Personal-org name is derived from the profile's **first + last name**: set once at profile registration (`user.repository.ts`) and used by the login hook and backfill; the signup placeholder is renamed once and then left stable on later edits.

### Schema & migration
- `packages/database/src/schema.ts`: extended `organizations` (`slug` unique, `logo`, `metadata`), added `organization_members` / `organization_invitations` / `teams` / `team_members`, and added `sessions.active_organization_id` / `active_team_id`.
- Single migration `0035_wild_moondragon.sql`: creates the org tables, adds the columns, drops `profiles.organization_id`, and cleans up the orphaned legacy institution-org rows (no slug + no members).

### Legacy institution-org removal
- Removed the free-text `organization` field from registration (`registration-form`/`registration-fields`), the profile API schema (`user.schema`), the profile UI, and the account-settings flow.
- Removed `profiles.organizationId`, `createOrReturnOrganization`, and the org-name anonymization from the backend.
- The profile now shows a **read-only** active-org chip via `authClient.useActiveOrganization()`.

### Tooling & tests
- Backfill script + `db:backfill:orgs` package script; seed provisions personal orgs for seeded users.
- New `personal-organization.spec.ts` (provisioning + naming, incl. idempotency and rename-on-registration); updated all specs that referenced the removed `organization` field; added the `organizations/` layer to the vitest sequencer.

### Deploy notes
- Backfill existing users by running two idempotent `INSERT … ON CONFLICT DO NOTHING` statements directly against the DB. Safe to re-run.


dominikvrbic:
On first look this looks good and the safe-to-ship reasoning holds. A few things I think we should fix before merge:

1. **Provisioning should be part of the migration, not a manual backfill.** Right now existing users only get a personal org via the hand-run `db:backfill:orgs` script or their next login. It's not wired into `db:migrate`, so it can be forgotten per environment and leaves a window where users have no org. The migration already does data changes (the legacy-org DELETE), so folding in the two idempotent INSERTs (the ones already in the deploy notes) makes provisioning automatic, atomic and consistent everywhere. If we keep the script, it should be a gated/automated deploy step with a post-run count check, not a manual note.

2. **Account deletion leaves the real name in the personal org.** `delete()` scrubs `profiles`/`users` but never touches `organizations`, so `organizations.name` keeps "<First> <Last>'s workspace", and since it's a soft-delete the member FK cascade never fires so the deleted user stays owner. We should anonymize/clean the personal org in the same transaction.

3. **The legacy-org DELETE has no test and is irreversible.** It's the riskiest SQL here. Let's add a test asserting it only removes the intended rows (and that backfill provisions existing profiled users), and consider archiving legacy names before the destructive delete.

4. **Profile-create + org-name sync aren't atomic.** In `createOrUpdateUserProfile` the profile insert and `syncPersonalOrganizationName` run as two separate statements. If the sync throws, the org name never gets corrected (the login path only names on create). Wrap them in a transaction.

Minor: `sessions.active_team_id` has no FK (add one to match `active_organization_id`); the backfill loops per-user where a set-based statement fits; PR body names the wrong migration file (`0032_harsh_unus` vs the actual `0032_worthless_doctor_octopus`).

dominikvrbic (approved):

PR #1779, opened 2026-07-13
feat(workbooks): inline command cell

dominikvrbic:
## Why

Long-term we are deprecating protocols as a separate concept: **a protocol is just a command**. Every past attempt to get there (#1659 enum cell, #1777 stacked reland, #1778 whole-product rename with DB migration) died from scope. This PR is the deliberate restart with the **smallest possible slice that gets a generic command into main and into real testing**.

The staged plan:

1. **This PR** — additive inline `command` cell (`string` | `json` | `yaml`), executable over serial and BLE exactly like a protocol, stored in the cell payload like a text/question block. No new table, no migration, old mobile apps skip it. Smallest reviewable unit.
2. **Validate** — command cells in real workbooks/flows on web + mobile against devices; confirm the free-form payload + `producerCellId` attribution model holds up.
3. **Kill protocols** — once validated, write the migration that removes protocols as a separate entity and migrates all existing protocols to commands (the #1778 work is the reference for scope: routes, DB rename, jsonb data-migration, carve-outs like the MQTT `{protocolId}` topic segment and `_protocol_set` firmware contract).

## What's included

- **@repo/api**: `zCommandCell` (`{format: string|json|yaml, content, name?}`), `command-payload` resolve/validate, `zMeasurementCommandContent` riding the existing `measurement` flow node (pre-command clients skip it instead of crashing), cells↔flow round-trip, known-command catalog (`device-command.schema.ts`).
- **Web**: command cell editor with format select, known-command autocomplete + hover hints (CodeMirror), execution via `useIotProtocolExecution.executeCommand`; graphical flow editor gets a `COMMAND` node (palette, flow-mapper round-trip, side-panel editor).
- **Mobile**: `CommandNode` runs inline commands through the scanner; scan attribution generalized from `protocolId` to `producerCellId` (persisted, wire-format pin updated) so downstream branch cells can read command output.
- **Backward compatible**: additive only — old mobile apps skip command cells; no schema/table changes.

## Provenance

Re-cut of the #1777 commits directly onto `main` (clean cherry-pick — the taxonomy stack dependency was unnecessary). Sander's known-command catalog from #1659 survives as autocomplete on a free-form editor instead of a closed enum. #1778's rename is intentionally excluded and becomes step 3.

## Verification (on this branch, based on `main`)

- `tsc` clean: @repo/api, web, mobile, backend
- @repo/api tests green (incl. schema, command-payload, cells↔flow round-trip)
- web: workbook components 263/263 + 17 new command completion/panel tests
- mobile: full suite green (wire-format pin updated for `producerCellId`)

Closes OJD-1556.

### DIFF
--- a/apps/mobile/package.json
+++ b/apps/mobile/package.json
@@ -115,6 +115,7 @@
     "tailwind-merge": "^3.5.0",
     "tailwindcss": "catalog:",
     "uuid": "^14.0.0",
+    "yaml": "2.8.3",
     "zod": "catalog:",
     "zustand": "^5.0.12"
   },
--- a/apps/mobile/src/features/measurement-flow/domain/flow-transitions.test.ts
+++ b/apps/mobile/src/features/measurement-flow/domain/flow-transitions.test.ts
@@ -20,6 +20,15 @@ const makeQuestion = (id: string): FlowNode =>
 const makeMeasurement = (id: string): FlowNode =>
   ({ id, type: "measurement", name: id, content: { params: {}, protocolId: "p" } }) as FlowNode;
 
+// A command cell also rides a "measurement" node but carries no protocolId.
+const makeCommand = (id: string): FlowNode =>
+  ({
+    id,
+    type: "measurement",
+    name: id,
+    content: { command: { format: "string", content: "battery" } },
+  }) as FlowNode;
+
 const inFlow = (overrides: Partial<FlowState> = {}): FlowState => ({
   ...initialFlowState,
   experimentId: "exp-1",
@@ -34,6 +43,14 @@ describe("flowProtocolId", () => {
 
   it("is undefined for questions-only flows", () => {
     expect(flowProtocolId([makeQuestion("q1"), makeQuestion("q2")])).toBeUndefined();
+  });
+
+  it("skips a leading command node so the real protocol still resolves", () => {
+    expect(flowProtocolId([makeCommand("c1"), makeMeasurement("m1")])).toBe("p");
+  });
+
+  it("is undefined for a command-only flow (no protocol to upload against)", () => {
+    expect(flowProtocolId([makeCommand("c1")])).toBeUndefined();
   });
 });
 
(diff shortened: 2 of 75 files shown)


vladstoe:
Findings after testing locally on mobile:

### 1. Command → Instruction → Branch → PASS/FAIL Markdown

- The Command executes successfully and returns the expected response.
- The same Branch selects the success path on web.
- On mobile, it selects the default/failure path because the Branch cannot resolve the Command’s `response` field.

### 2. Command → Protocol → Macro

- Both the Command and Protocol execute.
- Accepting the result does not create the local measurement.
- A comparison workbook using **Instruction → Protocol → Macro** accepts and creates the local measurement normally.
- The issue is therefore specific to having a Command in the workbook.

### 3. CommandNode i18n

- CommandNode contains broken or missing translations for the command step in mobile.

dominikvrbic:
Thanks for testing on device, Vlad. All three are fixed in 9b9865e:

**1. Branch can't resolve the Command's `response` on mobile (works on web).** Web wraps a scalar command result as `{ response: <value> }` (`toOutputData`), but mobile's `hydrateCells` kept the raw value, so the branch's `response` field resolved to `undefined` and fell through to the default path. `hydrateCells` now mirrors web for command producers: a plain object passes through; any scalar/array is wrapped as `{ response }`. Branch routing now matches web.

**2. Command → Protocol → Macro doesn't create the local measurement on accept.** A command rides a `measurement` flow node with no `protocolId`, and `flowProtocolId()` returned the first `measurement` node it found (the command) → `undefined`. `handleUploadMeasurement` then threw `Missing protocol id`, which the handler's `.catch` swallowed, so no measurement was created and the flow didn't advance. `flowProtocolId()` now skips command nodes and resolves the first node that actually has a `protocolId`. Instruction → Protocol → Macro worked because there was no command node to shadow the protocol.

**3. CommandNode i18n.** The `commandNode.*` keys were never added to the mobile `measurement-flow` locale, so the node rendered raw keys. Added them for en-US and nl-NL.

Regression tests added for (1) and (2): `hydrate-cells` (scalar→`response` wrap + object passthrough) and `flow-transitions` (`flowProtocolId` skips a leading command node). Mind re-testing the same three workbooks on device when you get a chance?

vladstoe (approved):

PR #1790, opened 2026-07-16
feat(backend): IAM Phase 2 — org-scoped authorization (shared access matrix, resource grants, declarative guards)

vladstoe:

- [x] Feature
- [ ] Bug Fix
- [x] Refactor
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Test Update
- [ ] Chore
- [ ] Other (please describe)

## Description

Second phase of the unified IAM / org-scoped access-control rollout. Phase 1 (#1751) added the org layer and provisioned a personal org per user. This phase makes resources **org-owned** and routes every access decision through a single authorization service, driven by **one Better Auth access-control matrix** shared by the auth server and the backend.

Behavior is intentionally **unchanged** for end users today: every org is a single-owner personal org, and public resources stay world-readable. This lands the machinery that later phases (sharing, teams, directory) build on.

## What's included

**One access-control matrix** (`packages/auth/src/access.ts`, exported as `@repo/auth/access`)
- Built with Better Auth's `createAccessControl` — the single source of truth for "which role may do what", shared by the Better Auth server config and the backend.
- `ResourceType = experiment | protocol | macro | workbook | device`; `ResourceAction = read | update | share | manage`.
- Roles: `owner` / `admin` = full control; `member` = read-only (folds the org base-permission baseline into the role for now; a per-org `base_permission` dial is deferred until multi-member orgs exist).
- `orgRoleCan` / `grantRoleCan` evaluate that matrix in-process via `role.authorize()`, so an **explicit** user's role can be checked without a request/session context (needed by the service layer and non-HTTP callers). Grant `viewer` maps to the read-only `member` role.

**Authorization service** (`apps/backend/src/authorization/`)
- `AuthorizationService.can(userId, { resourceType, resourceId, action })` — the single decision point. Precedence: **owning-org role** (`orgRoleCan`) → **per-resource grants** `user → team → org` (`grantRoleCan`) → **public-read** fallback.

**Declarative route guards** — authorization lives on the controller as decorators:
- `@CanAccess({ resource, action })` — per-resource check delegating to `can()`. Applied across **16 controllers**. Denials surface as **403 Forbidden**, except a resource that doesn't exist returns **404 Not Found**.
- `@CanCreateInOrg()` — membership check for creation into a target org (there is no resource id yet, so `can()` doesn't apply). **5 controllers.**
- `@CanContributeToExperiment()` — experiment-membership gate for data contribution, which read access alone does not grant. **2 controllers.**
- Read use-cases still keep an experiment-existence check so they return a correct not-found off the HTTP path (jobs, other use-cases, tests), independent of the route guard.

**Schema + migration** (`packages/database`)
- New `resource_grants` table + `resource_type` / `grantee_type` / `visibility` / `org_base_permission` enums.
- Added `organization_id` (nullable) and `visibility` to experiments / protocols / macros / workbooks / **iot devices**; added `base_permission` (default `read`) to organizations. Devices default to `private` visibility and use `ON DELETE RESTRICT` (they own a live AWS IoT Thing); other resources default `public` / `ON DELETE CASCADE`.
- Migration `0039_iam_org_scoped_access` with the data backfill **folded into the migration** (set-based + idempotent) so it runs automatically and atomically on `db:migrate` in every environment — no hand-run scripts. Each existing resource is assigned its creator's personal org (resolved via the `personal-<userId>` slug); experiment memberships are mirrored into `resource_grants` (role-preserving), and macros/protocols/workbooks/devices get a creator admin grant.

## Orphaned resources (known, intentional)

Resources created by a user with no personal org (e.g. deleted before Phase 1) keep `organization_id = NULL`. `can()` handles this safely — they stay world-readable via the `public` visibility default and are editable by no one, matching prior behavior. `NOT NULL` is intentionally **not** enforced yet; that depends on a deleted-user / orphan-handling decision owned by a later phase.

## Testing

- Unit specs for `AuthorizationService` and for each guard (`can-access`, `can-create-in-org`, `can-contribute-to-experiment`).
- **Per-controller authorization specs**: every guarded route asserts both a genuine denial (403) and that `can()` is invoked with the correct `{ resourceType, resourceId, action }` (catches wrong-action drift on a decorator).
- Use-case specs for join-request approval/rejection and member management covering the success, conflict, not-found, and repository/email-failure paths.
- Backfill migration covered by a mirror-SQL spec (`resource-org-backfill-migration.spec.ts`) and end-to-end verified on an isolated Postgres: seed pre-migration data → `db:migrate` → assert org assignment, role-preserving membership mirror, org-less-creator handling, and visibility defaults.

---
<a href="https://app.blacksmith.sh/Jan-IngenHousz-Institute/codesmith/open-jii/pr/1790"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786783646&installation_id=146274774&pr_number=1790&repository=Jan-IngenHousz-Institute%2Fopen-jii&return_to=https%3A%2F%2Fgithub.com%2FJan-IngenHousz-Institute%2Fopen-jii%2Fpull%2F1790&signature=5b298b2ab4cd9917134f746acea6f0b399de8b7365a42617fa33097dc328558e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>/codesmith</code> with what you need. Autofix is disabled.</sup>


dominikvrbic (commented):
## Authorization: use Better Auth org access control + guards/decorators, not `can()` inside the use-cases

The main thing I'd like to change before this lands: authorization shouldn't be hand-rolled and inlined in the use-cases.

Two things are hand-built here that the stack already gives us:

1. `abilities.ts` (`roleCan` / `basePermissionCan`) reinvents a role→action matrix.
2. Every mutating use-case inlines `authz.can(...)` + log + `failure(AppError.forbidden())`, copy-pasted across protocol/macro/workbook `update`/`delete` and the compat mutations. That couples the application layer to authz, and the first mutation that forgets to paste it is a silent hole.

Better Auth's organization plugin has a first-class access-control model for exactly this, and `@thallesp/nestjs-better-auth` (already our dependency) exposes it as a guard-backed decorator — see https://docs.nestjs.com/guards. Define the statement + roles once in `packages/auth/src/server.ts`:

```ts
import { createAccessControl } from "better-auth/plugins/access";

const statement = {
  experiment: ["read", "update", "share", "manage"],
  protocol:   ["read", "update", "share", "manage"],
  macro:      ["read", "update", "share", "manage"],
  workbook:   ["read", "update", "share", "manage"],
} as const;

const ac = createAccessControl(statement);
const member = ac.newRole({ /* read-only across types */ });
const admin  = ac.newRole({ /* full control */ });
const owner  = ac.newRole({ /* full control */ });

organization({ ac, roles: { owner, admin, member }, creatorRole: "owner", teams: { enabled: true } })
```

Then authorization is declared on the route and the use-cases go back to pure business logic:

```ts
@MemberHasPermission({ permissions: { protocol: ["update"] } })
@Implement(protocolContract.updateProtocol)
updateProtocol(@Session() session: UserSession) { /* no authz here */ }

@MemberHasPermission({ permissions: { protocol: ["manage"] } }) // delete = manage
@Implement(protocolContract.deleteProtocol)
deleteProtocol(@Session() session: UserSession) { /* ... */ }
```

This deletes `abilities.ts` outright, removes `AuthorizationService` from every use-case, and makes each route's requirement visible at a glance. Refs: [NestJS guards](https://docs.nestjs.com/guards) · [lib permission docs](https://github.com/ThallesP/nestjs-better-auth#permission-based-access-control).

**One thing to resolve before this can fully replace `can()`:** Better Auth AC is role→resource-type→action against the member's **active** org. It does not model (a) per-instance `resource_grants` (sharing one specific protocol with a user/team), (b) the resource's **owning** org vs the session's active org, or (c) the public/visibility read-fallback. So for Phase 2: if the role/permission layer is all we need right now, let's adopt AC + `@MemberHasPermission` and drop the bespoke matrix + `resource_grants` until a sharing phase actually needs it. If instance sharing is required now, keep a thin per-resource guard for *only* that dimension but still move the role matrix onto Better Auth AC. Also note `@MemberHasPermission` requires `activeOrganizationId` in the session, so we'd need to guarantee the personal org is set active.

---

**Secondary:** experiments still run on the old experiment-members RBAC while protocols/macros/workbooks go through this new path, so we now have two authorization systems in parallel. The PR flags this as intentional, but let's track consolidating experiments onto the same model so we don't maintain both indefinitely.

dominikvrbic:
Following up on the authorization point with what I found digging into the library.

**`@MemberHasPermission` on its own won't work for us.** Its guard hard-denies when there's no `activeOrganizationId`, and it only ever checks the **active** org (it calls `hasPermission` without an org id). Our resources are owned by a specific org that won't always be the active one, so the stock decorator would check the wrong org.

**But Better Auth's underlying API takes an explicit org.** In `plugins/organization/organization.mjs` the handler does:

```js
const activeOrganizationId = ctx.body.organizationId || ctx.context.session.session.activeOrganizationId;
```

So `auth.api.hasPermission({ headers, body: { organizationId: owningOrgId, permissions: { experiment: ["update"] } } })` resolves "is this user a member of *that* org and does their role there allow it" **live, server-side**. That's exactly what we need.

**Design I'm proposing** — custom decorator + guard, with the org-role tier delegated to Better Auth:

```ts
@CanAccess({ resource: "experiment", action: "update" })   // param defaults to "id"
@Implement(experimentContract.updateExperiment)
updateExperiment(@Session() s: UserSession) { /* pure handler, no authz */ }
```

The guard reads the metadata + `req.params.id` and calls `AuthorizationService.can()`. `can()` gets refactored so its org-role tier calls `auth.api.hasPermission({ organizationId: owningOrgId, ... })` instead of the hand-rolled matrix. That:

- **deletes `abilities.ts`** — Better Auth's `ac` + roles becomes the single permission matrix;
- checks the **owning** org, not the active one;
- keeps the two things Better Auth can't model — per-instance `resource_grants` (user -> team -> org) and public-read — as the remaining logic in `can()`;
- moves enforcement out of the use-cases and onto the route.

So `can()` survives as a thin orchestrator (Better Auth org check -> grants -> public-read), the bespoke matrix goes away, and authorization becomes declarative.

**No session extension.** We don't need to stuff all of a user's orgs into the session to make this work — the explicit-org `hasPermission` does a live lookup against the owning org. Denormalizing memberships into the session would only save that lookup and it would go **stale** (a removed or downgraded member stays authorized until their session refreshes), so let's not make the session an authorization source.

**Scope — everything moves onto this, no second system:**

- **Experiments:** drop the parallel experiment-members RBAC and put them on the same guard + grants model.
- **Devices:** add `organization_id` + `visibility` first (`loadOwnership` currently returns not-found for `device`), then they join.

End state: one authorization path for experiment / protocol / macro / workbook / device.

dominikvrbic:
Two updates, and a course-correction on my own earlier "no session extension" note.

**Platform decision: we're not using Better Auth's "active org" at all.** Authorization always runs against the resource's **owning** org — the guard passes `organizationId: owningOrgId` to `hasPermission`, never a session-level active org. Consequences:

- `@MemberHasPermission` is out (it hard-requires `activeOrganizationId` and only checks the active org).
- create can't use active org for the default owner; new resources default to the creator's personal org, with an optional explicit `organizationId` to create straight into a shared org.
- the `session.create.before` active-org hook (`server.ts` ~226) and the `activeOrganizationId` args threaded through the 4 create paths come out.

**Given active org is gone, a custom session is genuinely worth exploring** — and done right it doesn't have the staleness problem I flagged earlier. Better Auth's `customSession` plugin overrides `/get-session`: it loads the live session and runs your `fn` to augment it, so anything computed there is **fresh at fetch time**, not a persisted snapshot. So we could carry the user's full membership map `{ organizationId -> role }` on the session as the natural *replacement* for the single active-org field. That buys:

- a **fast path** for the org-role tier — the guard resolves the owning-org role from the session in memory: one membership query per request (at session fetch) instead of one per authz check;
- **UI data** for free (org switcher, owning-org badges) with no extra calls.

The distinction that matters: compute it in `customSession` (fresh per fetch) — *not* as a persisted/denormalized session column, which is the stale-until-refresh version I was warning against. `resource_grants` (user -> team -> org) and public-read stay per-resource and live; the session only ever accelerates the org-role tier. If we like it, this fits the no-active-org model better than Better Auth's built-in single-active-org ever did.

dominikvrbic:
Closing the loop on `createAccessControl` — confirming it **stays as the foundation**, since it's easy to read "drop active org + custom guard" as moving off Better Auth's access control. We don't.

`createAccessControl(statement)` + `ac.newRole(...)` in `packages/auth/src/server.ts` **is** the permission matrix (role -> `{resource: actions}`) — it's exactly what replaces the hand-rolled `abilities.ts` (`roleCan`/`basePermissionCan`). Two things to note:

- **It's what `hasPermission` evaluates against.** Our guard calls `auth.api.hasPermission({ organizationId: owningOrgId, permissions: { protocol: ["update"] } })`; Better Auth resolves the member's role in that org and checks it against these `ac`/roles. No `ac` -> nothing to evaluate.
- **It's required for our resource types.** The default org statement only covers `organization`/`member`/`invitation`/`team`. To authorize `experiment`/`protocol`/`macro`/`workbook`/`device` we have to declare them in a custom statement and wire `organization({ ac, roles })`. That's the missing piece in `server.ts` today — the org plugin is registered without `ac`/`roles`, so this is a prerequisite for the whole model.

It's orthogonal to dropping active org: `ac`/roles are org-plugin config, active org is a session field — we keep the former and drop the latter with no conflict. The only world where we *don't* use `createAccessControl` is if we kept `abilities.ts`, which adopting it is meant to delete.

dominikvrbic (approved):

PR #1791, opened 2026-07-16
feat: device identification handshake, global device context, Ambit/MiniPAR drivers, per-device dispatch

dominikvrbic:
## What

Multi-sensor workbooks need to know WHO they are talking to after a BLE/serial connect. This PR adds the whole global-device-context stack, one commit per slice:

1. **Identification handshake + \$device branch source** (OJD-1668, OJD-1658): \`identifyDevice()\` probes \`hello\` / \`{"command":"INFO"}\` and classifies against firmware-faithful signatures (MultispeQ \`MultispeQ Ready\` and legacy \`Instrument Ready\`, MiniPAR JSON/CSV, Ambit \`NEW ... Ready\` sentinel), returning a structured \`DeviceIdentity\` stored per connection on both hosts. The identified family outranks the toolbar selection (toast on mismatch). Branch cells gain the reserved \`\$device\` source (\`family/id/name/index/firmwareVersion\`) with a pinned "Connected device" entry in the condition editor. Adds the \`ambit\` sensor family (idempotent \`ADD VALUE\` migration; Ambit = leaf sensor, distinct from the Ambyte gateway that migration 0037 folded it into).
2. **ctx namespace** (OJD-1655): \`buildCellNamespace\` exposes every upstream output to macros as read-only \`ctx\` (by canonical name and \`byId\`), device-scoped for multi-device outputs, with \`ctx.\$device\` injected at run time. Threaded through the backend Lambda payload into all three sandbox wrappers (JS deep-freeze, Python MappingProxyType, R lockBinding) and both hosts, including mobile's Pyodide sandbox. \`resolveConditionValue\` delegates to the same resolver so branches and macros can never disagree. ctx-only macros run without a preceding measurement.
3. **Real Ambit + MiniPAR drivers** (OJD-1670, OJD-1671): AmbitDriver speaks the unframed text console (quiet-window collection, wake polling, silent writers with settle + hello re-verify, one-write multi-arg commands, parsed get_par/temp); protocol JSON is rejected with a command-cell hint. MiniParDriver is dual-mode (LINE with quiet-window fallback for newline-less replies, JSON protocol with the constant \`7A1E3AA1\` footer). Both register in \`identifyDevice\` and the web host; ambit/minipar become locally connectable (ambyte stays MQTT-only).
4. **Per-device dispatch** (OJD-1659): a \$device-conditioned branch is a dispatcher on both hosts: each connected device evaluates the branch with its own identity, devices group by resolved path, each path's protocol/command target runs against only its group, unmatched devices are skipped with a message, and consumed targets are skipped exactly once (no jump). Mobile rounds run per-device payloads via \`executeScanAssignments\` and upload per-result \`protocolId\` (own MQTT topic per family) under one shared \`workbook_run_id\`.

## Notes for review

- The widened name-uniqueness rule from the swartz prior art was deliberately NOT ported: \`zWorkbookCellArray\` validates loads too, so it would make existing workbooks with duplicate producer names unloadable. On ctx collisions the nearest upstream producer wins instead.
- Per-family mock devices: \`?mockDevices=multispeq,ambit,minipar\` (bare \`=1\` stays a MultispeQ); mocks answer the real handshake.
- Ambit firmware quirk: \`hello\` prints a hardcoded \`NEW Name Here\`, so classification keys on the sentinel, never the name.
- Macro-as-constructor, command-validator, calibration workbooks (OJD-1612/1613) and \`ctx.\$devices\` aggregates are out of scope.

## Verification

- packages/iot 313 tests, packages/api 675 tests, web 4703 tests (662 files), mobile 867 tests (82 files); lint + check-types green across all touched packages.
- Migration applied to local pg: enum extended, existing protocol rows untouched.
- Backend integration tests need the 5433 test DB (not running locally); backend typecheck green, execute-macro context threading covered by api schema tests.

Linear: OJD-1668, OJD-1658, OJD-1655, OJD-1670, OJD-1671, OJD-1659

Linear: [OJD-1680](https://linear.app/openjii/issue/OJD-1680/cloud-macro-execution-is-not-pinned-to-the-published-workbook-version)

### DIFF
--- a/apps/backend/src/macros/application/use-cases/execute-macro-batch/execute-macro-batch.spec.ts
+++ b/apps/backend/src/macros/application/use-cases/execute-macro-batch/execute-macro-batch.spec.ts
@@ -8,9 +8,14 @@ import {
   AppError,
 } from "../../../../common/utils/fp-utils";
 import { TestHarness } from "../../../../test/test-harness";
+import type { LambdaExecutionPayload } from "../../../core/models/macro-execution.model";
 import type { CreateMacroDto } from "../../../core/models/macro.model";
 import type { LambdaPort } from "../../../core/ports/lambda.port";
 import { LAMBDA_PORT } from "../../../core/ports/lambda.port";
+import {
+  macroSnapshotKey,
+  MacroSnapshotRepository,
+} from "../../../core/repositories/macro-snapshot.repository";
 import { MacroRepository } from "../../../core/repositories/macro.repository";
 import { ExecuteMacroBatchUseCase } from "./execute-macro-batch";
 
@@ -19,6 +24,7 @@ describe("ExecuteMacroBatchUseCase", () => {
   let testUserId: string;
   let useCase: ExecuteMacroBatchUseCase;
   let macroRepository: MacroRepository;
+  let macroSnapshotRepository: MacroSnapshotRepository;
   let lambdaPort: LambdaPort;
 
   beforeAll(async () => {
@@ -30,6 +36,7 @@ describe("ExecuteMacroBatchUseCase", () => {
     testUserId = await testApp.createTestUser({});
     useCase = testApp.module.get(ExecuteMacroBatchUseCase);
     macroRepository = testApp.module.get(MacroRepository);
+    macroSnapshotRepository = testApp.module.get(MacroSnapshotRepository);
     lambdaPort = testApp.module.get(LAMBDA_PORT);
   });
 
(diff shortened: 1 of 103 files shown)


dominikvrbic:
## Test scenarios

The following scenarios are covered by the automated suites and are the recommended acceptance pass for this PR.

### Device identification and selection

- Connect a current MultispeQ: `hello` identifies the family and `device_info` enriches name, ID, firmware, and battery.
- Connect legacy MultispeQ firmware returning `Instrument Ready`: identification still succeeds and falls back to `battery` + `hello` when `device_info` is unavailable.
- Connect MiniPAR in CSV and JSON hello modes: both resolve to `minipar`, including persisted name and firmware parsing.
- Connect Ambit: the `NEW ... Ready` sentinel resolves to `ambit`, while the firmware's hardcoded `NEW Name Here` text is not exposed as the device name.
- Choose a toolbar family that disagrees with the handshake: the identified family controls the driver and the UI reports the mismatch.
- Return an unknown handshake: the generic driver/fallback path remains usable.

### `$device` branching and per-device dispatch

- Select **Connected device** in a branch and evaluate `family`, `id`, `name`, `index`, and `firmwareVersion` fields.
- Change an existing condition to `$device`: invalid prior fields reset to `family`; valid device fields are preserved.
- Run a branch with multiple connected families: each device evaluates against its own identity and is grouped onto its matching path.
- Route different groups to different protocol/command targets: each target receives only its matching devices.
- Leave a device unmatched: it is skipped with a message rather than failing the run.
- Continue the linear workbook walk after dispatch: targets already consumed by the branch are skipped exactly once and are not run twice.
- Execute a device branch with no connections: the branch returns the expected no-device error output.

### Macro `ctx` namespace

- Read upstream question, protocol, and macro outputs through canonical names and `ctx.byId`.
- Confirm the nearest upstream producer wins when names collide and later/self outputs are unavailable.
- Run a ctx-only macro without a preceding measurement.
- For multiple devices, verify each macro receives only that device's measurement plus its own `ctx.$device` identity.
- Persist the primary macro result for downstream cells; empty and identical callbacks do not trigger repeated store writes.
- Confirm branch resolution and macro resolution use the same namespace semantics.

### Ambit and MiniPAR protocol behavior

- Ambit: exercise wake polling, idle re-wake, quiet-window collection, partial replies at the overall timeout, non-ASCII wake-byte stripping, overflow recovery, parsed PAR/temperature replies, bad commands, and silent calibration writers with hello re-verification.
- MiniPAR LINE mode: exercise newline and newline-less replies, error replies, persisted-name lookup, and overflow recovery.
- MiniPAR JSON mode: send the protocol in one write, reassemble chunked responses, validate the `7A1E3AA1` footer, retain inline firmware errors, and reject corrupt/incomplete envelopes.
- Confirm family-specific browser mock devices reproduce the relevant handshake, command, and protocol responses.

### Multi-device upload and cloud execution

- Upload per-device results with each result's own protocol ID/name under one shared workbook run ID.
- Include the published workbook version ID and each device's serialized macro context in the upload.
- Carry `workbook_version_id` and `macro_context` through Databricks bronze, silver, and gold processing.
- Group backend batch execution by `(macro_id, workbook_version_id)` and execute the immutable macro snapshot from that published workbook version.
- Edit the live macro after publication: cloud processing still runs the published snapshot, matching local/mobile execution.
- Process legacy records without a workbook version ID: they continue to use the live-macro fallback.
- Forward device-specific context into the sandbox for JavaScript, Python, and R execution.

### Verification result

- Full IoT coverage suite: 326 tests passed.
- Full mobile coverage suite: 874 tests passed.
- Web focused scenarios, lint, formatting, and type checks passed.
- GitHub `Build, Lint, & Test` passed.
- Codecov patch coverage: **95.89%** (target **93.65%**).

Related Linear work: OJD-1655, OJD-1658, OJD-1659, OJD-1668, OJD-1670, OJD-1671, OJD-1680.

vladstoe (approved):
seems to work fine, including the used version of a workbook for a measurement upload

dominikvrbic:
> seems to work fine, including the used version of a workbook for a measurement upload

then lets get it on dev and test more

PR #1802, opened 2026-07-20
feat(newsletter): Mailchimp newsletter subscription (footer, settings, registration)

dominikvrbic:
Implements [OJD-1393](https://linear.app/openjii/issue/OJD-1393/account-page-newsletter-opt-inunsubscribe-gdpr-blocker) (newsletter opt-in/unsubscribe, GDPR blocker). Unblocks OPENJII-16 / OPENJII-12.

## What

Mailchimp-backed newsletter subscription with **no local persistence** (Mailchimp is the source of truth), offered in three places:

- **Public footer** (home + info pages): anonymous email form, double opt-in (`pending` + Mailchimp confirmation email)
- **Account settings**: toggle with live Mailchimp status (subscribed / pending "check your inbox" / unsubscribed / provider-error)
- **Registration**: opt-in checkbox, unchecked by default; direct subscribe after signup (email just OTP-verified), fire-and-forget so it never blocks registration
- **Account deletion** permanently deletes the Mailchimp member (GDPR erasure)

## How

- New `newsletter` oRPC domain + NestJS Mailchimp module (config/adapter/port, mirrors the email/aws modules)
- Anonymous subscribe route is rate-limited (scoped `@nestjs/throttler`, keyed on the CloudFront-appended `X-Forwarded-For` entry, spoof-tested) and always returns a generic response (no email enumeration)
- Read-then-act subscribe: an existing subscriber re-entering their email in the footer is a no-op, never downgraded to pending
- Backend boots with zero `MAILCHIMP_*` env vars (feature dormant); strict zod validation once any are present
- "Community" group-vs-tag is pure config (`MAILCHIMP_COMMUNITY_KIND`)

## Verification

- backend 2164 tests / packages/api 1240 + typecheck / web lint + check-types + 57 focused tests, all green
- Three independent review passes (2x Codex, 1x Opus cohesive) - all findings fixed

## Operational verification

- [x] All six `MAILCHIMP_*` values configured in GitHub dev/prod environments; Terraform, CI, Secrets Manager, and ECS plumbing added. Applied successfully in dev; production resources will be created by the next production deploy.
- [x] Real Mailchimp end-to-end verification completed: local existing-member classification preserved `subscribed`; deployed dev anonymous subscription created `pending` with `MMERGE10=Community`. Disposable verification contacts were removed afterward.
- [x] Confirmed request chain: CloudFront appends the viewer IP and ALB's default append mode adds the CloudFront edge IP. Terraform does not override that ALB default; the guard selects the second-from-right entry and spoof-resistance is tested.
- [ ] Native review of de-DE copy. nl-NL remains disabled; its machine-authored copy must be reviewed before that locale is enabled.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -336,6 +336,12 @@ jobs:
       TF_VAR_contentful_access_token: ${{ secrets.CONTENTFUL_ACCESS_TOKEN }}
       TF_VAR_contentful_preview_access_token: ${{ secrets.CONTENTFUL_PREVIEW_ACCESS_TOKEN }}
       TF_VAR_contentful_preview_secret: ${{ secrets.CONTENTFUL_PREVIEW_SECRET }}
+      TF_VAR_mailchimp_api_key: ${{ secrets.MAILCHIMP_API_KEY }}
+      TF_VAR_mailchimp_server_prefix: ${{ secrets.MAILCHIMP_SERVER_PREFIX }}
+      TF_VAR_mailchimp_audience_id: ${{ secrets.MAILCHIMP_AUDIENCE_ID }}
+      TF_VAR_mailchimp_community_kind: ${{ secrets.MAILCHIMP_COMMUNITY_KIND }}
+      TF_VAR_mailchimp_community_id: ${{ secrets.MAILCHIMP_COMMUNITY_ID }}
+      TF_VAR_mailchimp_community_name: ${{ secrets.MAILCHIMP_COMMUNITY_NAME }}
       TF_VAR_api_cloudfront_header_value: ${{ secrets.API_CLOUDFRONT_HEADER_VALUE }}
       TF_VAR_orcid_oauth_client_id: ${{ secrets.ORCID_OAUTH_CLIENT_ID }}
       TF_VAR_orcid_oauth_client_secret: ${{ secrets.ORCID_OAUTH_CLIENT_SECRET }}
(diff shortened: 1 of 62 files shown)


vladstoe:
Did some e2e testing against real Mailchimp. The flow I ran into:

1. Created a fresh account → settings → toggled the newsletter on → worked, got subscribed ✅
2. Deleted that account
3. Re-registered with the **same email** + ticked the newsletter checkbox → checkbox errored, and later the settings toggle also failed:

```
400 "Forgotten Email Not Subscribed"
"<email> was permanently deleted and cannot be re-imported.
 The contact must re-subscribe to get back on the list."
```

Looks like account deletion permanently deletes the Mailchimp contact, and Mailchimp then blocks re-adding that email via the API (double opt-in fails too) — so a returning user with the same email can't re-subscribe from the app.

Is this expected? If it's the intended GDPR trade-off, no problem — but right now it surfaces as a generic 400 with no explanation, so I think we should catch this case and show a clearer message instead of a silent failure.

dominikvrbic:
@vladstoe Thanks — reproduced and confirmed. Good catch.

The permanent-delete behavior itself is intentional: account deletion calls Mailchimp's `delete-permanent` / “forget” action for GDPR erasure, not unsubscribe. Mailchimp then suppresses the address and rejects both direct and pending API re-adds with `400 Forgotten Email Not Subscribed`; only Mailchimp's hosted signup form can restore it.

The opaque failure is not intentional and is worth fixing. The adapter does not recognize this specific response, `toFailure` currently loses the domain error code, and Mailchimp's `title`/`detail` fields are not surfaced, so registration and settings collapse it to a generic error.

Recommendation: keep permanent erasure, add explicit forgotten-email detection/code propagation, and show clear guidance on the affected UI surfaces. This is not a data/security blocker, but it is a real UX gap and I recommend addressing it in this PR before merge.

vladstoe (approved):

Issue #16, opened 2025-01-06
Onboard Main Implementer (Developer Petar)


dominikvrbic:
## Description
- Provide Petar with access to the project tools and repositories.
- Walk through the PoC scope and objectives.
- Share relevant documentation, including the Milestone 1 deliverables.
- Align on key responsibilities, deadlines, and expected deliverables for the PoC.

## Acceptance Criteria
- Petar has access to all necessary project resources and documentation. (Github, AWS)
- A kick-off meeting is held with Petar to align on the PoC goals and tasks.
- Petar begins initial development work on the PoC.

PR #1852, opened 2026-07-30
feat: ojd 1638 rfc unified iam organization scoped access control phase 3

vladstoe:

- [x] Feature
- [x] Bug Fix
- [x] Refactor
- [x] Documentation Update
- [ ] Performance Improvement
- [ ] Test Update
- [ ] Chore
- [ ] Other (please describe)

---

Third phase of the IAM rollout ([#1751](https://github.com/Jan-IngenHousz-Institute/open-jii/pull/1751) Phase 1, [#1790](https://github.com/Jan-IngenHousz-Institute/open-jii/pull/1790) Phase 2). Phase 2 built the authorization substrate — `resource_grants`, `AuthorizationService.can()`, declarative guards — without changing behavior. This phase puts it to work:

- **Resources become shareable.** Experiments, macros, protocols, workbooks and IoT devices each get a Collaborators surface with two tiers.
- **Visibility becomes enforceable.** Macros, protocols and workbooks can be created private, publishing is one-way and enforced server-side, and a private resource is undiscoverable — including in global search.
- **Grants become the only thing that confers access.** The legacy `experiment_members` roster is retired from every runtime path: the grant *is* the access.

## Why

Experiments had two parallel access systems — the `experiment_members` roster (admin/member) driving the members UI, and the `resource_grants` layer that actually feeds authorization, kept in step by a runtime mirror. Building a sharing surface on top of both would have shipped two user-facing lists answering "who can touch this experiment." They collapse into one model here.

## The model

Two user-facing tiers, on every type:

- **Can view** — read. On experiments it also carries **contribute** (add measurements and annotations): with the contributor roster gone, a read-only tier would mean nothing on a public experiment, where everyone can already read.
- **Can edit** — everything: edit, manage, share, delete.

The baseline is unchanged: org **owner/admin** = full control of the org's resources, org **member** = read-only, public = read-only for everyone.

Access hangs on **ownership** rather than on a creator grant:

- Collaborators lists synthesize an **Owner** row from the resource's owning organization — a static badge with no tier dropdown, no revoke and no Leave. Owners are not collaborators.
- A creator gets no grant where ownership already covers it. The exception is creating into an organization you belong to as a plain `member`: membership is enough to create there, but `member` is read-only, so without a grant you could not manage what you just made.
- **The last-admin invariant is conditional.** While the owning org has a living owner, somebody is always in full control, so every grant is freely revocable. It bites only on a resource whose owning org has no living owner left, where the admin grants are all that remain.
- **Account deletion is blocked** while you are the sole living owner of an org that owns resources, or hold the sole admin grant on a resource whose owning org has no living owner. A bulk admin hand-off surface is the way out.

### Access model

- New `contribute` action, with the org-role and grant-role matrices split. Grant `viewer` → read + contribute (experiments), `admin` → all; org `member` → read-only.
- The roster-existence guard is gone: the five annotation routes, CSV upload and IoT `getUploadUrl` now gate on `can(contribute)`. **One exception inside upload**: appending to an existing table is `contribute`, but *creating* a new Databricks table still requires `manage`, as it did before. `contribute` means adding data to the experiment, not minting storage for it.
- **One name per tier — written, not yet stored.** The read-and-contribute tier was written as both `viewer` and `member` depending on the code path, with a normalization step papering over it — the same word the org matrix uses for a *read-only* role. `viewer` is now the only spelling **every writer emits**, and `GrantRole` in `@repo/database` types the write helpers off `string`, cross-checked against the matrix the way `STAFFING_GRANT_ROLES` already is. Readers deliberately still accept both: renaming stored rows is deferred to the follow-up release (see Migration), so during a rolling deploy an old pod can still write `member` and a new pod must read it as the same tier.
- **Grant inputs and outputs are deliberately asymmetric.** `zShareableRole` (`admin` | `viewer`) validates create/update payloads, so **`role: "owner"` now returns 400** where it previously succeeded — nothing writes `owner` and the UI cannot express it. Responses keep the three-value `zGrantRole` so a legacy `owner` row still deserializes.

### Sharing

- `listGrants` / `createGrant` / `updateGrant` / `revokeGrant`, plus organization-grantee search. Listing collaborators is gated on **share**, not read — no collaborator enumeration on a public resource.
- **Self-leave** is the one operation not gated on share: your own grant is the authority, which is what lets a "Can view" grantee remove themselves from a list they cannot see. Having no grant of your own answers with a uniform 404, disclosing nothing. It remains subject to the last-admin invariant, and is refused on an archived experiment.
- Every direct-grant write flows through one guarded path — staffing lock, last-admin invariant, archived-experiment refusal — so demote, revoke and self-leave are checked identically across all five types.
- Lifecycle hardening: no silent demotion via re-share, grantee validation without existence disclosure, transactional grant teardown on resource delete, atomic invitation claim, and re-authorization of the inviter at acceptance.

### Visibility & publishing

- Monotonic and server-enforced: `visibility` is removed from the experiment update body (it previously accepted public→private and persisted it with no transition check), `setVisibility` allows private→public only, and `embargoUntil` is editable only while private.
- Macros, protocols and workbooks gain optional `visibility` at create, a publish control and a visibility badge. Their `findAll` adopts experiment-style access scoping, and **global search inherits it**, so a private resource cannot be found by name.
- Publishing is confirmation-gated everywhere, because it cannot be undone. Experiments, macros and protocols expose it as a visibility dropdown that goes inert once public; the workbook shows visibility as a read-only field in its metadata row with a separate Publish action. Visibility badges appear on the list and overview surfaces of all five types.

### Experiments

- Membership endpoints and contracts are removed (`zExperimentMember`, add/remove member, `updateExperimentMemberRole`).
- Invitations are tier-only, and acceptance writes the grant. Join-request approval writes a viewer grant.
- `getExperimentAccess` returns a `capabilities` object, and every web gate moves off `createdBy` heuristics onto capabilities.
- Contributor credit derives from grantees and honours `anonymizeContributors`, so a flagged experiment serves the same pseudonyms the data grid uses.

### Devices

- Devices join the sharing model with the same Collaborators route, tiers, self-leave, staffing invariant and deletion blocker as the rest. "Can edit" on a device means rename, delete, and issue/rotate/revoke its AWS certificate.
- The detail page's six in-page tabs became routes, matching every other type. Credentials and the Overview danger zone are gated on `canManage`, and the credentials route redirects rather than rendering blank.
- Devices stay **unpublishable** — no visibility route, no publish control. That is enforced by the type rather than a runtime refusal, so widening the sharing enum cannot silently make a device publishable.
- The surface ships behind the existing `IOT_DEVICES` flag.

### Web UI

- **Unified Collaborators surface.** The experiment page keeps its shape, with the Members tab becoming Collaborators: grant rows, two tiers, user or organization grantees, one invite modal. Every tab mirrors the capability its endpoint is guarded on — Collaborators and Invited on share, **Requests on manage**, which previously rendered for everyone so a viewer got a Requests tab whose 403 read as "no requests".
- On macros, protocols, workbooks and devices, **Collaborators is a route rather than an in-page tab**, with the strip in the layout under the title. Switching tabs swaps the whole view, and the tab is linkable and back-button-able.
- **Workbook Delete moved off the list row.** Scoped lists legitimately contain public and viewer-shared workbooks, and rows carry no capabilities (they are detail-only, to avoid a `can()` per row), so a row could not tell a manager from a reader and offered a Delete that could only 403. It now sits in the Overview action row beside Publish and Fork, gated on `canManage` — the previous danger-zone section is gone.
- **Destructive actions on experiments follow the same rule.** Experiment Delete was gated on the `EXPERIMENT_DELETION` flag alone, so with the flag on a read-only viewer was offered a Delete button the route would refuse; it is gated on `canManage` now, alongside Archive. The flag still governs whether *managers* may delete at all, never who may.
- **Viewer self-leave**: a "Your access" card with a Leave action for grantees who cannot see the list, using the same confirmation as the row-based self-revoke.
- Session hygiene: authorization-sensitive caches are principal-scoped and dropped on sign-out, so switching accounts in one browser cannot flash the previous user's capabilities, collaborator names or invitee emails.
- First-load correctness: the access query stays disabled until the session resolves, and now reports that as *loading* rather than *no data* — previously every experiment route painted "Experiment not found" before its content. The same class of fix applies to the collaborators list (skeleton rows instead of a premature empty state) and to self-demotion, which now downgrades the UI immediately instead of 403ing until reload.
- i18n: member-role vocabulary retired and the `sharing` copy rewritten — **and added to de-DE and nl-NL, which were missing it entirely**. The new visibility copy is en-US only so far, so those two locales fall back to English for it.
- Docs: the collaborators, sharing and visibility guides are rewritten for the two-tier model, and `openapi.json` is regenerated.
- **Drive-by fix**, unrelated to IAM: the new-macro form showed a language-specific starter template that was never part of the form value, so submitting without touching it failed with "Code file content is required". The template is now seeded into form state and follows the language select while the field is untouched.

## What will feel different

| Before | Now |
| --- | --- |
| A Members tab with admin/member roles, and no sharing surface anywhere | A Collaborators tab of grant rows with two tiers ("Can view" / "Can edit"), user **or organization** grantees, and one invite modal |
| Only experiments could be shared, and only by granting full membership | All five types are shareable — with no read-only tier on experiments, since "Can view" carries contribute |
| The member roster, real names included, was listable by anyone with read — strangers on a public experiment included | Collaborator lists are gated on share; everyone else sees contributor credit only |
| Leaving removed your roster row | Leaving revokes your grant: "Can edit" users from their own row, "Can view" users from a "Your access" card. An owner holds no grant and so has no Leave — they render as a static **Owner** row |
| Macros, protocols and workbooks were always public, with no sharing or visibility surface | Private at create, undiscoverable while private, confirmation-gated one-way publish, and Overview / Collaborators routes |
| Only experiments had collaborators or blocked account deletion; macros, protocols, workbooks and devices had neither | All five behave alike: an Owner row heads the list, the last admin grant is protected once no living owner remains, and a solely-owned resource blocks account deletion until it is handed over |
| Devices were owner-private, with no way to give anyone else access | Devices are shareable, with both tiers, self-leave, and a place in the deletion blocker (badged **Device**, falling back to the serial number when unnamed). Still unpublishable |
| Collaborators lived in an in-page tab on macros/protocols/workbooks — switching it left the details sidebar and content on screen, and the tab had no URL | A route on every type: the view swaps wholesale, and the tab is linkable |

## Migration (`0041_iam_grants_sole_access_source.sql`)

**This is release 1 of a two-release expand/contract.** The migration deploys before the new backend, and old pods stay live during the rollout — so anything that renames stored roles has to wait. One column default (`resource_grants.role` moves to `'viewer'`) and two set-based data statements:

1. **Backfill**: every `experiment_members` row gets its matching grant where one is absent, with the roster's `member` written as the grant vocabulary's `viewer`. This closes a real gap — invitation acceptance used to write a roster row *without* a mirror grant, so those users would have silently lost access the moment the roster stopped being read.
2. **Cleanup**: strip creators' own grants across all five types, so no creator is a collaborator on their own resource. It is guarded on the creator actually holding `owner`/`admin` in the owning org, which is what makes it provably access-neutral: rows predating the org backfill still carry a NULL `organization_id`, where the creator's grant is their only access, and a creator who holds only `member` keeps theirs because `member` is read-only.

Both statements are access-neutral in **both** directions — the previous release already reads `viewer` grants — which is what makes them safe while two versions serve.

### Deferred to release 2 (follow-up PR, after this is fully rolled out)

The `member` → `viewer` rewrites and the `invitations.role` default. Both were unsafe here, in opposite directions:

- An old pod accepting a **rewritten invitation** tries to insert `viewer` into the legacy `experiment_members.role` enum and fails. Acceptance lives inside the sign-in hook, which swallows failures and exits early once `registered` flips — so the invitation is lost permanently, with the inviter left looking at "pending".
- An old pod writing `role='member'` **after** a one-shot rewrite produces a row the new `zGrantRole` enum rejects, which breaks the whole Collaborators panel for that resource.

So release 1 keeps both spellings readable and has invitation creation keep writing `member` for the low tier — the one place an old pod cannot tolerate the new word. Release 2 renames the stored rows, flips invitation writes to `viewer`, and removes the read tolerance. Release 3 from the review folds into 2: release-1 pods never write `member` grants, and `normaliseTier` stays permanently as a totality guard.

Every statement is idempotent, and the rebuilt file was validated on a scratch database — migrate to 0040, seed every branch of both statements, apply 0041 — then applied to a populated local database with identical before/after row counts.

`experiment_members` itself is left physically untouched — schema identical to main, data frozen, read only by the migration, and its own `member` roster role deliberately not renamed. Removing the table is a separate decision.

## Testing

- Backend, api and web suites green, with typecheck and lint across auth/api/backend/web/i18n/docs.
- Every load-bearing guarantee carries a red/green proof: removing either lock, the closed-account create guard, the deletion tombstone or the multi-role token matching makes the corresponding test fail.
- Concurrency is covered by cross-connection race specs for the staffing lock and the atomic invitation claim.
- The migration is verified by a mirror-SQL spec, and the release-1 file was validated on a scratch database (migrate to 0040, seed every branch of both statements, apply 0041 — 11 SQL assertions) before being applied to a populated local database: 242 grants and 28 invitations of existing data, preserved exactly. Every statement is idempotent — re-applying the whole file changes nothing and the cleanup removes zero rows. No role is renamed in this release, so no tier count moves.
- An end-to-end suite of **77 scenarios** drives the real stack against a real browser and database — all green. It lives in a **standalone repo**, not in this changeset, which is why it is not reviewable from the diff. Device rows are seeded by SQL rather than through registration, so no test creates a real AWS IoT Thing, and no test asserts a *successful* certificate operation — only the refusal, which fails at the authorization guard before any AWS call.

| Spec | Area | n | Covers |
| --- | --- | --- | --- |
| **A** | Layout — experiments | 5 | One tab strip, one filter, user search only inside Invite, no people tabs without capability |
| **B** | Sharing — experiments | 13 | Both tiers end to end, org grantees, public-vs-granted contribution, revoke, live self-demotion |
| **C** | Invitations — experiments | 6 | Email invite lifecycle, registered-address and duplicate handling, signup claims the grant, revoked invite grants nothing |
| **D** | Join requests — experiments | 3 | Request visibility, approval writes a Can view grant, rejection adds no access |
| **E** | Staffing — all five | 8 | Owner is not demotable or leavable, last-admin race under true concurrency, deletion blockers across all types, hand-off clears them |
| **F** | Visibility — all four publishable | 6 | One-way publish on all four types, embargo editable only while private, private child of a public parent stays hidden and says so to a collaborator |
| **G** | Sharing — macros, protocols, workbooks | 8 | Collaborators as a *route*, owner holds no grant, private resource undiscoverable in list/search/direct URL, workbook Delete placement |
| **H** | Contributors — experiments | 3 | Credit from user grants at every tier, stable pseudonym for strangers, anonymization leaves the gated list alone |
| **I** | Session cache — experiments | 3 | No cached share-gated identities across a session switch, no not-found flash on first load |
| **J** | Archive — experiments | 4 | Archived surface read-only, disabled leave card, access loss closes an open invite form on refetch |
| **K** | Sharing — devices | 8 | Owner row, share-gated list, every credential operation refused for Can view, stranger can neither list nor open |
| **L** | Mixed-version rollout | 2 | A grant stored under the retired `member` spelling reads, contributes and renders as Can view; promoting it stores the current spelling |
| **M** | Review regressions | 8 | One row per reviewed finding — cross-user deletion, path/payload id contradiction, cross-workbook version, archived invitation, join-request state across an account switch, table creation vs appending, single-decision claim, reader join action |

<details>
<summary>All 77 scenarios</summary>

**A — layout, experiments** · A1 one collaborators/invited/requests strip and no sharing card · A2 exactly one page filter; user search only inside Invite · A3 Invite offers User/Organization and two tiers with Can view default · A4 the top filter drives counts for grants, invitations and requests · A5 a reader with no capability gets no people tabs and no gated fetch

**B — sharing, experiments** · B1 Ana grants Ben Can edit and Vic Can view · B2 Vic can read the now-shared private experiment · B3 Can view passes annotation and upload authorization · B4 Can view has no update/manage controls · B5 Can edit updates metadata and can open management surfaces · B6 Vic edit surfaces follow Can edit then disappear after downgrade · B7 revoking Vic removes list and direct access · B8 TeamLab Can view lets Omar read and contribute · B9 Sam can read a public experiment · B10 public access alone cannot annotate or upload · B11 public hint explains contribution; Sam Can view can annotate · B12 unrelated Omar cannot open Ana's personal private experiment · B13 Ben demoting himself loses the management surface without a reload

**C — invitations, experiments** · C1 new email invitation shows address, tier, pending badge and revoke · C2 exact registered email offers the account, not send-by-email · C3 existing collaborator email explains that access already exists · C4 invited-email signup via DB OTP accepts the viewer grant · C5 revoked invitation does not grant access on later signup · C6 viewer has no Invited tab and never calls listInvitations

**D — join requests, experiments** · D1 Sam request with message is visible to Ana and Ben · D2 approving yields Sam a Can view collaborator grant and contribution · D3 rejection clears request without adding access

**E — staffing, all five types** · E1 the owner has no tier control to demote herself with · E2 the owner has nothing to revoke or leave · E3 a second admin can leave after a warning · E4 granting another admin does not make the owner leavable · E5 account deletion is blocked by every sole-admin resource, not just experiments · E6 true simultaneous last-admin race · E7 a Can view grantee gives up their own access from the leave card · E8 hand-off clears deletion blockers and the account can be deleted

**F — visibility, experiments + macros/protocols/workbooks** · F1 private experiment offers a visibility dropdown and no way back · F2 embargo date is editable while private · F3 publishing is final: the dropdown goes inert and says why · F4 macro, protocol and workbook share the one-way publish behavior · F5 a public experiment's private workbook is invisible to an outside reader · F6 an experiment collaborator with no workbook access is told so on the design tab

**G — sharing, macros/protocols/workbooks** · G1 private macro uses Overview and Collaborators tabs, and they are routes · G2 resource Collaborators route adds Vic without invitation/request tabs · G2b the protocol owner holds no grant, and a granted admin is freely removable · G3 a Can view grantee gets a Collaborators route holding only the leave card · G4 stranger cannot discover a private macro in list, global search, or direct URL · G5 Can edit tier unlocks editing for a non-creator · G6 workbook Delete lives in the manager detail action row, not the list row · G7 Can edit unlocks the workbook's distinct update surface

**H — contributors, experiments** · H1 overview credit trail comes from user grants at every tier · H2 stranger sees one stable six-hex pseudonym in contributor credit and data grid · H3 anonymization does not alter Ben's share-gated collaborator list

**I — session cache, experiments** · I1 Ana to Vic switch never paints cached share-gated identities or fires their requests · I2 Vic to Ana switch restores the full share-capable view · I3 an experiment's first load never paints not-found while the session resolves

**J — archive, experiments** · J1 archived collaborator surface keeps one strip and makes every action read-only · J1b an archived Can view grantee sees a disabled leave card · J2 initial collaborator flow states and grants Can view · J3 loss of access closes an already-open invite form on focus refetch

**K — sharing, devices** · K1 the device detail page is a route-linked strip with a Collaborators tab · K2 the owner renders as an Owner row and can share the device Can view · K3 a Can view grantee sees the device but is refused every credential operation · K4 the collaborators list is share-gated, and the grantee's tab holds only their exit · K5 Can edit unlocks the certificate surface for a non-owner · K6 revoking takes the device back out of the grantee's registry · K7 a grantee can give up their own access from the Collaborators route · K8 a stranger can neither list nor open the device

**L — mixed-version rollout** · L1 a grant stored under the retired `member` spelling reads, contributes and renders as Can view · L2 promoting such a row through the picker stores the current spelling

**M — review regressions** · M1 one account cannot delete another, and the target survives · M2 a path id contradicted by the payload is rejected, not silently swapped · M3 a version of another workbook is not served through a readable one · M4 an invitation accepted after archival grants nothing · M5 an account switch never leaves the previous user's join request on screen · M6 contributing does not carry table creation, but still carries appending · M7 a join request can only be decided once · M8 a reader reaches a join action on Collaborators, not an empty page

</details>

## Phase 4

The organization product surface: org management, teams and the member directory, team grantees, and resource transfer — which is what gives an owner a way to hand over what they own, and what makes leaving an *organization*, rather than a single resource, possible.


dominikvrbic (changes requested):
## Request changes — consolidated IAM Phase 3 review

> **Review attribution:** This AI-assisted review was prepared at @dominikvrbic's request by Codex, with independent **GPT-5.6 Sol (max reasoning)** and **Claude Opus (max reasoning)** review passes. Findings were deduplicated and revalidated against PR head `bff5e8d3ebf5893c1dcc5ce5e9e8b0171b3d479b`.

I am requesting changes because the current head contains two critical authorization failures, several rollout/invariant failures, and concurrency holes that can disclose private data, mutate a different resource than the one authorized, delete another user's account, or permanently orphan resources.

### P0 — critical

#### 1. `@CanAccess` authorizes the path ID, but oRPC can execute against a client-overridden ID

**Locations:** `apps/backend/src/authorization/can-access.guard.ts:75-112`; representative handlers in `apps/backend/src/experiments/presentation/experiment.controller.ts:79-88` and `experiment-data-annotations.controller.ts:25-39`; pinned `@orpc/openapi@1.14.6` compact input codec.

The guard reads `request.params`, while the default compact oRPC codec later assembles input as `{ ...params, ...data }`, so query/body data wins. A caller can authorize against accessible resource A and make the handler operate on inaccessible resource B:

- `GET /api/v1/experiments/<A>?id=<B>` reads B after the guard approved A.
- `POST /api/v1/experiments/<A>/data/annotations` with body `{"id":"<B>", ...}` writes to B after approving A.

This affects the authorized surface broadly. The merge mechanism predates this PR, but this PR makes `@CanAccess` the central gate and expands the private-resource blast radius to macros, protocols, workbooks, devices, and newly reclassified actions.

**Required fix:** make route parameters non-overridable (for example detailed input structure, reapply flattened path params last, or pass the guard-authorized ID to handlers). Add real HTTP tests with mismatched path/query and path/body IDs for both read and write routes.

#### 2. Any authenticated user can delete another user's account

**Locations:** `apps/backend/src/users/presentation/user.controller.ts:31-46`; `apps/backend/src/users/application/use-cases/delete-user/delete-user.ts:22-69`. Compare the ownership check already present in `user.controller.ts:48-57`.

`deleteUser` accepts `input.id` without reading the session or comparing the target to the caller. `DeleteUserUseCase` receives only that ID, so there is no downstream ownership check. Any authenticated account can therefore call `DELETE /api/v1/users/<victimId>` and delete another user whenever the victim passes deletion blockers. The PR increases the destructive impact by removing the victim's grant graph and writing the deletion tombstone used by staffing invariants.

`user.controller.spec.ts:413-442` currently codifies the bug by authenticating as one user and deleting another.

**Required fix:** require `input.id === session.user.id` (or remove the path-selected target entirely), return forbidden for cross-user requests, and replace that test with explicit self-delete and cross-user rejection cases.

### P1 — high

#### 3. `getWorkbookVersion` authorizes one workbook and serves a version from another

**Locations:** `apps/backend/src/workbooks/presentation/workbook.controller.ts:160-164`; `get-workbook-version/get-workbook-version.ts:14-15`; `workbook-version.repository.ts:30-40`.

The guard authorizes path workbook A, but the use case loads `versionId` without constraining it to A. A caller who can read any workbook can request a version belonging to a private workbook and receive its cells and entity snapshots. This becomes a private-content disclosure because this PR introduces private workbooks.

**Required fix:** query by both `workbookId` and `versionId` and return not found on a mismatched pair. Add a real use-case/HTTP regression test.

#### 4. `setWorkbookVersion` can capture private workbook versions after the caller's access is revoked

**Locations:** `apps/backend/src/experiments/application/use-cases/set-workbook-version/set-workbook-version.ts:32-73`; compare the new check in `upgrade-workbook-version.ts:55-70`.

This PR correctly hardened `upgradeWorkbookVersion` against a revoked grantee capturing later workbook state, but its twin `setWorkbookVersion` still checks only management of the experiment. A former workbook grantee can select a newer version and materialize its cells into an experiment they manage.

**Required fix:** apply the same `can(workbook, read)` check before loading/setting the target version and add the equivalent revoked-grantee test.

#### 5. Migration `0041` is unsafe during the repository's migration-before-backend rolling rollout

**Locations:** `packages/database/drizzle/0041_iam_grants_sole_access_source.sql:36-41,60-68,122-131`; `packages/auth/src/access.ts:174-189`; `.github/workflows/deploy.yml:118-158`.

The migration rewrites invitation/grant roles from `member` to `viewer` while old pods can remain live. Old code accepting a rewritten invitation attempts to insert `viewer` into the legacy `experiment_members.role` enum (`admin|member`) and fails. Conversely, an old pod can write a new `resource_grants.role='member'` after the one-shot rewrite; new code no longer recognizes that role, so access silently disappears and collaborator output can fail validation.

The workflow can also deploy backend code while migrations are skipped: production change detection may compare only `HEAD~1`, and `deploy-backend` accepts the migration job's `skipped` result.

**Required fix:** use an expand/contract rollout across releases: first accept both spellings and canonicalize/dual-write, deploy and drain old tasks, then rewrite data; remove legacy support only in a later release. Alternatively, enforce a real quiesced cutover. Do not rely on the current one-shot migration/order alone.

#### 6. Removing the legacy experiment-member API breaks the backend-before-frontend compatibility window

**Locations:** deleted `packages/api/src/domains/experiment/members/experiment-members.contract.ts`; replacement `packages/api/src/domains/sharing/sharing.contract.ts:27-75`; `.github/workflows/deploy.yml:137-173`.

The backend is deployed before the new web bundle. During that window, the previously served bundle—and already-open/cached clients afterward—continues calling the removed list/add/update/remove/transfer member routes and receives 404s, breaking collaborator management and account-deletion handoff.

**Required fix:** retain the old endpoints for one release as adapters to resource grants, deploy the new frontend, then remove the aliases in a later release.

#### 7. The deletion-handoff endpoint is an unrestricted archived-experiment sharing bypass

**Locations:** `packages/api/src/domains/sharing/transfer-admin/sharing-transfer-admin.schema.ts:5-22`; `apps/backend/src/sharing/application/use-cases/transfer-resource-admin/transfer-resource-admin.ts:92-124`; `apps/backend/src/sharing/core/repositories/sharing.repository.ts:381-395`.

The contract describes a caller who is the current sole admin and is blocked from deleting their account, but the implementation proves only `can(share)` plus target selectability and then deliberately bypasses the archived-resource guard. Any owner/admin can use it to grant or promote an arbitrary selectable user to direct admin on a private archived experiment. Existing tests exercise and bless this bypass without proving the caller is deletion-blocked.

**Required fix:** in one transaction, lock and prove that `(caller, resource)` is a current sole-admin deletion blocker before invoking the exceptional write. All other calls must use the normal guarded sharing path and reject archived resources.

#### 8. A pending invitation accepted after archival creates a collaborator who cannot be removed

**Locations:** `apps/backend/src/users/application/use-cases/accept-pending-invitations/accept-pending-invitations.ts:60-108`; `apps/backend/src/users/core/repositories/user-invitation.repository.ts:241-285`; archived guard in `apps/backend/src/sharing/core/repositories/sharing.repository.ts:56-80`.

Acceptance reauthorizes with `can(share)`, which does not inspect experiment status, then upserts the canonical grant without the archived check. An invitation issued while active can therefore be accepted after archival. The viewer gains archived read/contribute access, while normal revoke and self-leave are blocked because the resource is archived.

**Required fix:** claim the invitation and lock/re-read the experiment row and status in the same transaction before the grant upsert. Retire/revoke the invitation if archived, and make all grant-producing paths contend with archive on the same resource lock.

#### 9. Self-deactivation lets the final admin bypass the staffing invariant and orphan a resource

**Locations:** `apps/backend/src/sharing/core/resource-staffing.ts:360-374,430-464`; `apps/backend/src/authorization/authorization.service.ts:55-149`; `packages/api/src/domains/user/user.schema.ts:42-48`.

Staffing treats only `activated=true` grantees as administrable, and waives last-admin protection when the target is not administrable. Authorization, however, ignores `activated`. A sole direct admin can deactivate themselves and, using the still-authorized session, leave, revoke, or demote their own grant. A resource—especially a device with live credentials—can then have no living owner/admin and become permanently unmanageable.

**Required fix:** align staffing and authorization. The smallest safe change is to base grant administrability on `deleted_at` rather than the self-healing `activated` flag; alternatively, make deactivated accounts unable to authorize. Add self-deactivate plus leave/revoke/demote concurrency tests.

### P2 — medium

#### 10. `createGrant` can commit an orphan grant after concurrent resource deletion

**Locations:** `apps/backend/src/sharing/application/use-cases/create-grant/create-grant.ts:34-58`; `apps/backend/src/sharing/core/repositories/sharing.repository.ts:56-80,258-311`; `packages/database/src/resource-grants.ts:86-103`.

Authorization/existence is checked before the write transaction. The transaction does not require the concrete resource to exist (and several resource types receive no check), while polymorphic grants have no resource FK. Deletion can sweep grants and delete the resource between those steps, after which `createGrant` inserts a permanent orphan and returns success.

**Fix:** lock the concrete resource row inside the grant transaction, require existence/unarchived status, and make deletion contend on that same row.

#### 11. Grant creation and admin handoff race target-account deletion

**Locations:** `create-grant.ts:43-58`; `transfer-resource-admin.ts:106-119`; `sharing.repository.ts:258-311,381-395`; `apps/backend/src/users/core/repositories/user.repository.ts:424-475`.

Target selectability is checked before the write transaction. A target can be closed and have grants swept before the later insert/promotion, leaving a grant to a deleted account because there is no grantee FK.

**Fix:** acquire `lockUserAccount(tx, targetUserId)`, recheck active/not-deleted inside the transaction, then write; add equivalent locked existence checks for organization grantees.

#### 12. Join-request approval can demote a concurrently promoted admin to viewer

**Locations:** `apps/backend/src/experiments/application/use-cases/experiment-join-requests/approve-join-request.ts:61-89`; `apps/backend/src/experiments/core/repositories/experiment-join-request.repository.ts:142-179`.

The “already collaborator” check is outside the write transaction, followed by an unconditional viewer upsert. If another administrator promotes the requester after the check, approval overwrites the newer admin grant with viewer.

**Fix:** atomically claim a pending request and make the grant write monotonic (`ON CONFLICT DO NOTHING`, or update only when the existing tier is lower).

#### 13. CSV upload now lets every Can-view collaborator create unbounded Databricks tables

**Locations:** `apps/backend/src/experiments/presentation/experiment-data-uploads.controller.ts:37`; `apps/backend/src/experiments/core/repositories/experiment-data-uploads.repository.ts:43-75`.

The route changed from `manage` to `contribute`. With `targetKind:'new'`, it creates a new table in the experiment's Databricks schema and triggers a job, with no per-experiment cap. Because viewer grants confer `contribute`, a reader—including every member of a granted team/org—can mint unbounded tables.

**Fix:** keep `contribute` for appending to an existing upload table, require `update`/`manage` for `targetKind:'new'`, and add a table/resource cap.

#### 14. Contributor pseudonyms are recomputable from identifiers exposed to readers

**Locations:** `apps/backend/src/experiments/application/use-cases/list-experiment-contributors/list-experiment-contributors.ts:35-48`; `contributor-anonymizer.service.ts:19-26`; creator metadata projection in `apps/backend/src/experiments/core/repositories/experiment.repository.ts:473-476`.

The pseudonym is the first six hex characters of `sha256(experimentId:userId)`. The experiment ID is the URL, creator/user IDs are available through related responses/search, and the data grid uses the same mapping. Readers can build a dictionary and bind “anonymous” contributor rows back to real identities.

**Fix:** use an HMAC with a server-held key or a per-experiment random salt, and suppress identifying creator metadata for callers without sharing authority when anonymization is enabled.

#### 15. Migration `0041` can remove the only staffing grant from an organization with no living owner

**Locations:** `packages/database/drizzle/0041_iam_grants_sole_access_source.sql:100-121`; `apps/backend/src/sharing/core/resource-staffing.ts:256-288`; `apps/backend/src/users/core/repositories/user.repository.ts:78-140`.

Runtime seeding omits a creator grant only when the organization both confers full control and has a living owner. Migration statement 2 checks only the role. If the sole owner is soft-deleted and an org admin creator holds the backfilled admin grant, the migration deletes that grant, weakening deletion blockers and allowing the resource to become unstaffed.

**Fix:** mirror the runtime living-owner condition exactly in the migration and add a fixture with a soft-deleted sole owner.

#### 16. `getDeletionBlockers` cache leaks names and candidate identities across account switches

**Locations:** `apps/web/hooks/profile/useDeletionBlockers/useDeletionBlockers.ts:5-12`; `apps/web/hooks/auth/useSignOut/useSignOut.ts:28-36`; `apps/web/hooks/sharing/resource-cache-keys.ts:11-25`; module-level client in `apps/web/providers/QueryProvider.tsx:11`.

This caller-sensitive query is not principal-scoped and is absent from the sign-out sweep. The module-level QueryClient survives sign-out, so the next user can briefly receive the previous user's private blocker resource names and candidate user metadata from cache.

**Fix:** principal-scope the key and clear it on sign-out. Given multiple omissions in the allowlist, prefer `queryClient.clear()` followed by re-seeding the signed-out session state.

#### 17. The previous user's pending join request is rendered after an account switch

**Locations:** `apps/web/hooks/experiment/join-request/useMyJoinRequest/useMyJoinRequest.ts:10-23`; `apps/web/hooks/auth/useSignOut/useSignOut.ts:20-36`; `apps/web/components/experiment-settings/collaborators/experiment-request-to-join.tsx:41-87`.

The query key contains only the experiment ID and is absent from the sign-out sweep. User B can therefore see user A's pending-request state and receive a live cancel button after switching accounts in the same tab. The server correctly refuses cancellation by B, but the cross-account disclosure and broken UI remain.

**Fix:** principal-scope the query and clear it on sign-out; cover an A-sign-out/B-sign-in transition.

#### 18. Read-only users are routed to a dead-end Collaborators page and lost the roster they previously had

**Locations:** `apps/web/app/[locale]/platform/experiments/[id]/layout.tsx:133-137`; `.../collaborators/experiment-collaborators-content.tsx:33-63,119-148,211-216`; `apps/web/components/experiment-overview/experiment-details/experiment-details-card.tsx:167-173`.

For an org member reading a private experiment or a public reader, `canShare`, `canManage`, and `canLeave` are all false, so the page renders only its heading/description/docs link. The application nevertheless exposes the Collaborators tab and “N collaborators” link. Pre-PR, readers could see the roster; the replacement contributors endpoint is not used here.

**Fix:** render the read-gated contributors list for this capability combination, or hide both entry points when the page has no reader-facing content. Add a read-only render case.

### Validation and scope notes

- All findings above were checked against exact PR head `bff5e8d3`; no repository files or Git state were changed during review.
- The independent web verification passed: **5,168 tests passed, 2 skipped, across 714 files**, and `tsc --noEmit` was clean. `git diff --check` also passed.
- The locale audit passed: the `sharing.*` key sets match across en-US, de-DE, and nl-NL, with the documented English fallback for `resourceVisibility.*`.
- The two failing production image scan checks report dependency CVEs (`brace-expansion` CVE-2026-69152 and `ip-address` CVE-2026-69192), separate from the authorization findings above.
- The PR describes a 67-scenario browser/database suite, but those Playwright specs/configuration are not present in this changeset, so that evidence could not be reproduced. This matters particularly for mixed-version rollout and cross-account cache-isolation behavior.

Please address the P0/P1 items before merge and add regression coverage for the mismatched-resource-ID, cross-user deletion, mixed-version deployment, archived invitation, and staffing/concurrency scenarios. The P2 items should be resolved in this PR where they arise from the new access-control model; if any are intentionally deferred, please record the accepted behavior and follow-up owner explicitly.

vladstoe:
Thanks — genuinely useful pass. All nine P0/P1 and seven of nine P2 are closed across four commits.

## P0 / P1

**1 — the payload could override the authorized path id.** The guard now **400s when the two disagree**, rather than reordering input as you suggested: the generated client strips path params before sending, so no legitimate request can carry a conflict, and making params win would mean rewriting `req.originalUrl`.

**2 — anyone could delete anyone's account.** Self-only check as the handler's first statement, **before** the existence lookup so 403-vs-404 can't probe which ids are real.

**3 — a version could be served from another workbook.** Lookup constrained by both ids as you specified, with `workbookId` required in the repository so it can't be omitted later.

**4 — a revoked grantee could still pin a new version.** Took the twin's `can(workbook, read)` check, as you specified.

**5 — the migration was unsafe mid-rollout.** Your expand/contract, **release 1 here**: rename deferred, both spellings read, invitations keep the old word. Release 2 follows once this is rolled out; your releases 2 and 3 fold together safely.

**6 — removing the member API breaks the deploy window.** Window accepted, **no adapters**: honouring the new invariants changes their behaviour so the stale bundle still breaks, and bypassing them reopens #7–#9. Real exposure is a tab left open across a deploy — 404s on one panel, no writes, fixed by reload.

**7 — the deletion hand-off skipped the archived guard.** The exemption now has to be **earned**: sole-admin-blocker proof for that resource, under lock, inside the method itself. Kept the exemption — routing through `guardedWrite` would strand anyone whose blocker *is* an archived experiment.

**8 — an invitation accepted after archival was unremovable.** Status re-read `FOR UPDATE` inside the acceptance transaction; archived stamps the invitation `revoked` and applies nothing.

**9 — a deactivated sole admin could orphan a resource.** Split the flag that was doing two jobs: **inertness** waives the invariant, **adequacy** is what gets counted. Not your primary suggestion (dropping `activated`) — that also lets an admin leave beside a deactivated one; not the alternative either, which would lock out accounts whose grants are seeded by invitation before onboarding.

## P2

**10 — a grant could land on a resource being deleted.** Both sides claim the row now; your *"make deletion contend on that same row"* half turned out to be load-bearing rather than optional, since every delete sweeps the grants before deleting the row. One departure: the existence check is **not** in the shared guarded write as you asked — self-leave's uniform 404 depends on tolerating a missing resource — so it sits on the create path only, with a test pinning that leaving a deleted resource still answers identically to leaving an unknown id.

**11 — a grant could land on a deleted account.** Lock and recheck inside the transaction as you specified, plus the org-grantee equivalent.

**12 — approval could demote a concurrently promoted admin.** Both halves as specified — atomic claim, monotonic write. The claim also closed a second defect you didn't flag: the status update carried no pending predicate, so two admins approving the same request both committed, giving two decisions and two emails.

**13 — Can-view could create unbounded tables.** Creating a table requires `manage` again; appending stays `contribute`. `manage` rather than your `update` because that is literally what `main` required. **The cap is not fixed** — pre-existing, and with creation back behind `manage` its reachable audience matches `main`.

**14 — pseudonyms are recomputable.** **Deferred** — it doesn't arise from the new access-control model: `contributor-anonymizer.service.ts` has zero diff against `main` and the formula dates to #1469 (2026-05-27), so this phase neither introduces nor widens it. Accepted behaviour: on a flag-on experiment, any authenticated user can enumerate the directory, recompute the formula and recover identities — as on `main` today. Tracked as its own follow-up rather than stretched into this PR.

**15 — the migration could strip an org's only staffing grant.** Accepted, not fixed. The scenario needs two different people in one organization, and there are none — every org is personal and single-member. Mirroring the runtime condition would instead *preserve* closed accounts' creator grants, which is worse than the current behaviour.

**16 — deletion-blocker cache leaked across accounts.** Added to the sign-out sweep, key left alone: it already contains the caller's id, so this was residency after sign-out rather than cross-account disclosure. Your **secondary** suggestion — clear the client outright — is the better fix and is queued with the `QueryProvider` redesign instead of adding a fourth allowlist entry to forget.

**17 — the previous user's join request rendered after a switch.** Principal-scoped key plus session gating, on the pattern the sweep already established next door. Unlike #16, this one really was key-collidable.

**18 — readers hit a dead-end Collaborators page.** Neither of your two remedies: the tab stays and the roster stays share-gated, but a reader now gets the **Request to join** action there instead of nothing — it already existed one page away with nothing pointing at it. Whether readers should see the roster is still open.

dominikvrbic (approved):

Issue #1855, opened 2026-08-01
Prod: MultispeQ measurements stuck unprocessed / marked Error in BELIS_1st_measurement_Rep1 (2026-08-01)


dominikvrbic:
## Report

Reported by **Maria Mina** ([mariamina@itqb.unl.pt](<mailto:mariamina@itqb.unl.pt>), ITQB NOVA) by email on **2026-08-01**.

> Today, I was in the field measuring my plants and then I went to the lab to open the JII software in my computer and I realized that, from 9:47:40 am (answering the question "measurement:143"), until 12:31:24 (answering the question "measurement: 34") the raw data is not processed. So I have raw data from all measurements. From our side, we did the measurements always in the same way.

## Affected

|  |  |
| -- | -- |
| Experiment | `BELIS_1st_measurement_Rep1` |
| Experiment ID | `6b821e4c-c325-4a75-9b35-aed382d3c80b` |
| URL | [https://openjii.org/en-US/platform/experiments/6b821e4c-c325-4a75-9b35-aed382d3c80b/data](<https://openjii.org/en-US/platform/experiments/6b821e4c-c325-4a75-9b35-aed382d3c80b/data>) |
| Device | MultispeQ |
| Timezone | Europe/Lisbon |
| Window | 2026-08-01, ~09:47:40 -> ~12:31:24 local |
| Reporter | Maria Mina - [mariamina@itqb.unl.pt](<mailto:mariamina@itqb.unl.pt>) |
| Collaborator | Moses Oyeleye Taiwo - [moses.taiwo@itqb.unl.pt](<mailto:moses.taiwo@itqb.unl.pt>) |

## Symptom

In the experiment Data table, rows in the affected window show a red **Error** status. Those rows still carry the raw fields that arrive with the upload (device type, timezone, the `measurement` question answer, sample id, latitude, longitude) but the **processed/derived columns are empty** - the values that appear for the healthy rows immediately above (e.g. the two trailing numeric columns) are blank on the Error rows.

So: raw ingest looks fine, the processing/enrichment step is what failed. The user still has the raw data.

Screenshot from the reporter shows healthy rows (status `none`) with full values, then a run of rows with status `Error` and truncated values, starting around measurement 34 and continuing into 16.

## Impact

A full morning of field measurements for a real research experiment has no processed output. The data is not lost (raw is present) so a reprocess/backfill should be able to recover it, but that path needs to be confirmed.

## Investigation needed

[...]

PR #1857, opened 2026-08-01
fix(data): block unsafe macro table full refresh

dominikvrbic:
## Summary

- prevent resets of the `experiment_macro_data` streaming table at the DLT layer
- reject the table in the manual full-refresh workflow
- replace the unsafe workflow example and add regression coverage

## Why

A full refresh would re-execute 241,610 unpinned historical macro rows with current code. This is ticket 01 of OJD-1702 and intentionally ships alone before the remaining hardening work.

## Verification

- `uv run pytest tests/test_macro_refresh_safety.py`
- `uv run pytest -m "not spark"`
- `uv run ruff check tests/test_macro_refresh_safety.py`
- YAML parse and `git diff --check`

Refs OJD-1702 / GH #1855.

### DIFF
--- a/.github/workflows/dlt-full-refresh.yml
+++ b/.github/workflows/dlt-full-refresh.yml
@@ -15,7 +15,7 @@ on:
           - prod
         required: true
       tables:
-        description: 'Comma-separated tables to full-refresh (e.g. "experiment_macro_data,experiment_macro_data_sandbox")'
+        description: 'Comma-separated tables to full-refresh (e.g. "experiment_status,experiment_table_metadata")'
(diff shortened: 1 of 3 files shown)


dominikvrbic:
Closing unmerged. We've decided to run the `experiment_macro_data` full refresh on prod to recover the OJD-1702 errored rows, so the guard this PR adds would block the exact operation we need. The branch stays pushed if we want to revisit a narrower guard later (e.g. require an explicit confirmation input rather than a hard block).

Refs OJD-1702 / #1855.

PR #1864, opened 2026-08-05
test(e2e): port the Playwright harness to @playwright/test

dominikvrbic:
Replaces the ad-hoc Python Playwright scripts with a dedicated `@playwright/test` workspace. The local setup commands it uses landed in #1863.

## Why `apps/e2e`

Keeping Playwright outside `apps/web` prevents Vitest's default glob from collecting browser specs. The package is named `@repo/e2e` so release preview treats it as an internal workspace, and its browser script is named `e2e` so `turbo test` never starts it by accident.

## What changed

- Added a Playwright setup project that signs in through the real email form, reads a fresh OTP from local Postgres, and writes new storage state for each run. Setup failures retain the same screenshots, video, and retry trace as test failures.
- Added focused coverage for navigation shortcuts, activity and account navigation, experiment chrome, server-backed workbook search states, and workbook attachment.
- Added an automatic browser-error guard. It ignores browser-cancelled requests and only allows specific status codes from integrations disabled by the local environment.
- Ported the maintenance recorder to TypeScript and removed the Python script.
- Added a nightly Web E2E workflow with manual dispatch.

## CI design

CI builds `backend`, `web`, and their workspace dependencies through Turbo, then prepares the Next standalone static and public assets. Playwright owns both production server processes, waits for the backend health endpoint and the rendered `/en-US/login` page, and stops the processes after the run.

Failed runs upload the HTML report, test results, traces, videos, screenshots, and both server logs. Workflow concurrency cancels superseded runs on the same branch.

The suite uses a production web build in CI to remove per-route compilation from browser timing. Local runs target an already-running `pnpm dev:fb` stack.

## Reliability and safety

- Fixture assertions only compare E2E-owned workbooks, so unrelated seed data can match the same search without failing the test.
- Experiment access uses `resource_grants`; the suite no longer depends on the dormant `experiment_members` table.
- Fixture cleanup refuses to mutate anything except the `openjii_local` database on a loopback host unless the caller sets `E2E_ALLOW_UNSAFE_DATABASE=1`.
- Keyboard shortcuts use Playwright's cross-platform `ControlOrMeta` modifier.
- The experiment name is a semantic level-one heading, which gives tests a stable user-facing locator without adding a production-only test hook.
- Optional firmware repositories stay unset in the generated local environment, matching backend validation and allowing production startup.

## Verification

- E2E ESLint and TypeScript checks pass.
- Devkit lint, type checks, tests, and generated-environment consistency checks pass.
- Playwright discovers the setup project and seven runnable Chromium tests, plus one tracked `fixme` for the existing direct account-page hydration bug.
- A Turbo production build of `backend`, `web`, and all required workspace packages passes.
- The focused experiment-title test passes.
- Changed files pass Prettier and `git diff --check`.
- The manually dispatched [Web E2E run](https://github.com/Jan-IngenHousz-Institute/open-jii/actions/runs/32861528985) passed: 8 tests passed and the tracked `fixme` was skipped.


petar-INFO (commented):
The harness architecture is genuinely good and has already been used productively, but **both specs fail against current main today**, verified by running them live. All three causes are specific and fixable.

Some notes from my side:

## Blocking (each with its fix)

1. **`Control+b` fails on every Mac.** Main moved the sidebar toggle out of the shadcn component (which accepted Ctrl *or* Cmd) into TanStack hotkeys as `Mod+B` (`apps/web/components/shortcuts/shortcuts-root.tsx`), and `Mod` is platform-aware: Cmd on macOS. The spec presses `Control+b` and times out waiting for the sidebar state to change. Green in Linux CI, red for every local Mac run, which is exactly the audience of the README's local-run instructions. Fix: `ControlOrMeta+b`.
2. **Closed-world assertions over a database the spec only partially owns.** `search("chloro")` asserts exact equality with a one-element array; current main's seed added `[Seed] Wheat Phenotyping Workbook`, which matches: not by name, but because workbook search also matches **linked protocol names**, and its pinned protocol is `[Seed] SPAD Chlorophyll Index`. So even relational seed data the spec never touches can enter the result set. The fixtures are properly tagged (`metadata->>'e2e'`) but the assertions ignore the tag. Fix: assert tagged fixtures present and tagged foils absent; never assert result-list length or completeness.
3. **`experiment_members` is dormant and the harness leans on it.** `findSeedExperimentId()` in `apps/e2e/helpers.ts` joins `experiment_members`; current main's seed *deletes* rows there and writes none (access control moved to `resource_grants`; main's own seed comment calls the table dormant). After rebase the helper throws `No active [Seed] experiment found`. `workbook-fixtures.ts` also inserts into it (dead weight). Fix: select by `experiments.created_by` + name prefix; drop the insert.

## Should fix in this PR

- **Add a nightly `schedule` to the workflow.** It is `workflow_dispatch` only, and the three blockers above are the cost made visible: the suite was green when written and rotted silently across ~22 commits of main. A suite nobody runs is red precisely when someone finally needs it.
- **Upload server logs on failure.** The workflow writes `/tmp/openjii-*.log` but uploads only `playwright-report/` and `test-results/`. Two lines in the artifact paths spare someone a blind CI debugging session.
- **First-character typing flake in `global-setup.ts`.** The email goes in via `pressSequentially`, which was observed dropping the leading character (symptom: an inexplicable OTP timeout, the OTP row lands under a truncated identifier). Use `fill()` plus an `inputValue()` assertion.
- **Brittle selectors in the chrome spec** (`.text-2xl`, `xpath=..`) contradict the branch's own skill guidance ("user-visible, deterministic"). Give the experiment title a role or test id.
- **The known-bug detour.** "Client-side navigation avoids the known direct-load account-page hydration failure" routes a smoke test around a real hydration bug. Link an issue in that comment or add a `test.fixme`, so the workaround cannot quietly become permanent.

## Non-blocking

- **Feature-flag override is the biggest capability gap**: without a PostHog key every flag resolves to its hardcoded default, so nothing behind `iot-devices` (default `false`) is testable end to end. Any future flag-gated feature inherits the same blind spot. Fine as a follow-up rather than blocking this PR.
- Two deep specs beat ten shallow ones. The `browserErrorGuard` fixture (auto-failing tests on unexpected console errors, with `expected-noise.ts` allowlisting exactly the known-dead local integrations) is the pattern that catches hydration bugs jsdom cannot. The response-synchronized search assertions (waiting for the actual `?search=` response instead of sleeping) are the right way to test debounced search. CI using a production build, with the reason documented, is the right call.
- The Python recorder deletion (`apps/web/e2e/record_maintenance.py`) alongside its TS replacement is appreciated; no debris left behind.

## Follow-ups surfaced by real use (not requested of this PR)

Building walkthrough recordings on this harness surfaced four gaps. We plan to pick these up in a follow-up branch once this lands; noting them here so they trace back to this review.

- **A fresh, empty seed user.** The seed identity owns experiments and workbooks, so first-run, empty-state and onboarding surfaces cannot be exercised against it. A second seeded user with no data, provisioned through `ensurePersonalOrganization` (a real invariant: a user without a personal org breaks resource creation), covers that class.
- **A per-user state reset helper.** Iterating on any per-user-state feature currently means hand-written `UPDATE profiles ...` between runs. Something like `pnpm --filter @repo/devkit reset-user -- --email ...`, or an `apps/e2e` fixture.
- **A right-server preflight in `global-setup.ts`.** If the target port is already held by a server from another checkout, the suite happily tests the wrong code. Observed in practice; it cost a full debugging cycle. Failing setup when the listener's cwd is not this checkout closes it.
- **Promote recording instrumentation into `helpers.ts`.** `record-maintenance.ts` proves the video pattern, but the pieces that found real bugs during recording work were console/pageerror capture, HTTP >= 400 logging to stdout, and login-as-arbitrary-user. Sharing those makes the next recording or spec ten lines instead of a page.

dominikvrbic:
Addressed all current blocking and should-fix points from Petar's review in the rebased stack:\n\n- use ControlOrMeta for the sidebar shortcut\n- make workbook assertions fixture-scoped and open-world\n- remove experiment_members as an access dependency and seed the current resource grant\n- add nightly scheduling and server-log artifacts\n- replace sequential typing with fill plus value verification\n- replace CSS/XPath title selectors with a stable test id\n- track the direct-load hydration bug with test.fixme\n- align the E2E tsx dependency with the workspace after CI caught the version mismatch\n\nFocused web tests, E2E lint/types, Playwright discovery, workspace dependency lint, Prettier, and diff checks pass locally. Fresh CI is running on the final stacked head.

dominikvrbic:
Follow-up: the rebased stack head is now `d05370164`. The full CI suite is green, including the database-backed **Build, Lint, & Test** job. That run confirms the unique partial-name fixture fixes the cross-suite collision seen on the prior run. All review threads on both stack PRs are resolved.

petar-INFO (approved):

dominikvrbic:
@petar-INFO Resolved. I did a second review of the scripts and CI beyond the original checklist and tightened the parts that were still weak:

- Playwright authentication is now a setup project with normal retry traces, screenshots, and video.
- CI builds backend, web, and their workspace dependencies through Turbo, then lets Playwright own server startup, readiness, and shutdown.
- The workflow is nightly and manually dispatchable, cancels superseded branch runs, and retains both server logs with Playwright artifacts on failure.
- The browser-error guard ignores only explicit browser cancellations and exact status codes from the local integrations we deliberately disable. Other console, page, request, and HTTP failures still fail the test.
- Database fixture cleanup is restricted to loopback `openjii_local` unless an explicit unsafe override is set.
- The long specs were split into focused cases, fixed sleeps and coordinate clicks were removed, and the production test-id hook was replaced with a semantic level-one heading.
- The generated local environment no longer supplies invalid firmware repository placeholders that prevent the backend from starting.

The manually dispatched [Web E2E run](https://github.com/Jan-IngenHousz-Institute/open-jii/actions/runs/32861528985) is green: 8 passed and the tracked direct-load hydration `fixme` was skipped. All PR checks are also green, including Build, Lint, & Test and the production backend/web builds. Head is `c39eee1ed`.

PR #1858, opened 2026-08-01
feat(macros): add failure alerting and shape diagnostics

dominikvrbic:
## Stack

* Depends on Jan-IngenHousz-Institute/open-jii#1857 (ticket 01 reset guard)
* This PR contains tickets 02 and 03 of Jan-IngenHousz-Institute/open-jii#1855

## Summary

* extract macro batch failures from the exact backend use-case log and alert through the existing Grafana/Slack path
* log value-free input-shape fingerprints on both sides of the backend/sandbox boundary
* keep JavaScript, Python, and R diagnostics canonical through shared fixtures and real-container tests
* add the cross-language contract to affected macro-sandbox CI

## Privacy

Fingerprints contain structure, permitted set labels, and macro/workbook identifiers only. Measurement values and GPS coordinates are never logged.

## Verification

* OpenTofu plan test and mutation audit for A/B/C + Slack routing
* table-driven alert replay tests
* backend specs and production build
* real JavaScript/Python/R container contract: 24/24
* shared backend fingerprint fixtures: 7/7
* wrapper diagnostics: 2/2
* focused lint/format/syntax checks and `git diff --check`
* independent GPT-5.6 Sol review: no remaining findings

Refs Jan-IngenHousz-Institute/open-jii#1855 / GH Jan-IngenHousz-Institute/open-jii#1855.

PR #1877, opened 2026-08-08
feat(macros): add failure alerting and shape diagnostics

dominikvrbic:
Replaces #1858, which could not be retargeted off the closed #1857 stack. Rebased onto `main` to drop `7696c336c` (the full-refresh guard rejected with #1857).

## Summary

* extract macro batch failures from the backend use-case log and alert through the existing Grafana/Slack path
* log input-shape fingerprints on both sides of the backend/sandbox boundary, so a recurrence of OJD-1702 is diagnosable
* carry sandbox fingerprints on the wrapper's structured stdout envelope, never stderr

**User-facing behaviour change:** JavaScript macro errors now include the error type — `TypeError: bad input` instead of `bad input`. This string propagates into the downstream `macro_error` column, so anything parsing those messages sees the new format.

## Privacy

Fingerprints carry JSON structure, not values. Two deliberate constraints, both added after independent review found the original approach unsound:

* **stderr is untrusted.** Macro code shares the wrapper's stderr stream and can write to it (a JS escape via `Object.constructor("return process")()` was demonstrated). Handlers therefore no longer forward wrapper stderr on success; it surfaces only when structured output fails to parse. Fingerprints travel in the parsed stdout envelope and are stripped before the Lambda response, so they never reach `macro_error`.
* **keys and labels are user-controlled.** The request schema accepts an arbitrary record, so a coordinate can arrive as a key name. Known structural keys and labels matching `^[A-Za-z0-9_-]{1,24}$` pass verbatim; everything else becomes a non-reversible FNV-1a digest, identical across TypeScript, JavaScript, Python and R.

**Scope of the claim:** measurement values and coordinates are not emitted. This is not an unconditional guarantee that no user-authored string ever appears: a short token-shaped label still passes verbatim by design, because label identity is the diagnostic payload.

## Diagnosability

`setTypeof` distinguishes `.set` being an object, string or null. `setIsArray` alone collapsed all three into one signature, and "`.set` was not an array" is precisely the OJD-1702 signature. Batch fingerprints are aggregated per distinct shape with a bounded `sampleItemIds`, so a 5000-item batch emits a handful of records instead of 5000 while offending rows stay identifiable.

## Verification

* shared fixture is v2 with 10 cases, including the three `.set`-not-an-array shapes the incident produced and a coordinate-as-key counterexample; it previously had no case for either
* all 10 cases produce byte-identical fingerprints in the real JavaScript, Python and R wrapper containers, with no privacy sentinel present on success or all-item-failure paths
* fixture parity for `wrapper.js`, `wrapper.py` and `wrapper.R` is now an automated vitest suite (spawns the real wrappers; skips cleanly when numpy/pandas/scipy or R+jsonlite are absent, e.g. CI)
* the demonstrated stderr escape no longer reaches container logs, and `fingerprints` does not appear in any Lambda response
* OpenTofu plan test and mutation audit for A/B/C + Slack routing; backend specs

## Known gaps

* The container contract suite was removed at review request, so the handler stderr non-forwarding fix has **no automated regression test**. It was verified once by execution. A refactor could reopen it silently.
* CodeRabbit flagged `FILL(m1, 0)` on the new alert query. Not changed: the same pattern is used 6 times on `main` in alerts already running in production, so changing only the new one would make it inconsistent with its siblings.
* CodeRabbit flagged the metric filter `$.context` as never matching. Not changed: the use-case constructs `new Logger(ExecuteMacroBatchUseCase.name)`, which is exactly what the filter expects.
* The `macro_batch_failure_threshold` of 10 has not been validated against real production failure rates.

Refs OJD-1702 / #1855.


dominikvrbic:
Removed the `macro-sandbox-contract` CI job and its tests per review.

Dropped:
- the `Cross-Language Contract (Macro Sandbox)` job in `.github/workflows/production-checks.yml` (file now identical to `main`)
- the fingerprint parity/privacy assertions this PR added to `test/handler-container.spec.ts` (file now identical to `main`)
- commit `754d0da20` (`fix(ci): build shared Vitest config before sandbox contract`), which existed only to make that job pass

Kept, because they are not part of that job:
- `packages/api/fixtures/macro-input-shape-fingerprints.json` and its `package.json` export, still consumed by `macro-execution.model.spec.ts` on the backend
- `test/wrapper.spec.ts`, which runs under the normal `test:run` suite, not `test:container`

Now 21 files, +992/-12 (was 23 / +1099). No new CI jobs.

PR #1879, opened 2026-08-10
feat(iot): drop protocolId from the ingest topic, keep the legacy shape as a transitional channel

petar-INFO:

- [x] Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Test Update
- [ ] Chore
- [ ] Other (please describe)

## Description

Removes the trailing `{protocolId}` segment from the MQTT ingest topic. The lean shape is now the canonical channel; the old 8-segment shape stays as a clearly marked transitional channel until the mobile fleet rolls over.

```
experiment/data_ingest/v1/{experimentId}/{sensorType}/{sensorVersion}/{sensorId}
```

### Why

- The segment carried no routing or security value: nothing subscribes per protocol, and the IoT policy renders it as a wildcard.
- Nothing consumes it downstream: gold and enriched pass the column through, no join, filter, or grouping anywhere.
- It was extracted **wrong** the whole time: the silver pipeline read segment 4 (the sensor family) and labeled it `protocol_id`, so the warehouse column contains values like `ambyte` instead of a protocol id for every MQTT row. The true id sits unparsed in the stored raw `topic` column and is backfillable.
- New procedure types (inline commands) have no protocolId, so the old contract could not even name their topic.

Protocol attribution belongs in the payload or is derivable from `workbook_version_id` (already sent). The payload schema is deliberately untouched by this PR.

### Changes

- **asyncapi.yaml**: ingest channel redefined without `{protocolId}`; the legacy 8-segment shape kept as a transitional channel (delete after mobile rollover). `x-infra-name` on the lean channel because both now share a static prefix.
- **infrastructure/modules/iot-core/main.tf**: rule and policy names honor an optional `x-infra-name` channel extension. Existing channels keep their derived names, so no live resource is renamed or replaced (policies are attached to live identities).
- **silver clean_data.py**: `protocol_id` is now taken from the legacy topic's trailing segment only (correct position), and is null on the lean shape. Fixes the wrong-segment extraction.
- **ADR** `mqtt-topic-proposal.mdx`: superseding note.

## OJD-1602

### DIFF
--- a/apps/data/src/pipelines/centrum/silver/clean_data.py
+++ b/apps/data/src/pipelines/centrum/silver/clean_data.py
@@ -70,10 +70,16 @@ def clean_data():
         )
         .withColumn("output", F.col("parsed_data.output"))
         .withColumn("user_id", F.col("parsed_data.user_id"))
-        # Which protocol produced the row; topic = experiment/data_ingest/v1/<experiment>/<protocol>/...
+        # Which protocol produced the row. Only the legacy 8-segment topic
+        # (.../{sensorId}/{protocolId}) carries it, as its trailing segment; the
+        # lean 7-segment shape has no protocol in the topic. The previous
+        # extraction read segment 4, which is the sensor family, not a protocol.
         .withColumn(
             "protocol_id",
-            F.expr(r"nullif(regexp_extract(parsed_data.topic, 'experiment/data_ingest/v1/[^/]+/([^/]+)', 1), '')")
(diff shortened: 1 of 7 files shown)


dominikvrbic (approved):

dominikvrbic:
Shall I modify the ambyte right away, and do you want to add to your pr the mobile change?

PR #1900, opened 2026-08-13
feat(iot): device connectivity, live fleet-index state, lifecycle-event history & last-data monitoring

petar-INFO:

- [x] Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Test Update
- [ ] Chore
- [ ] Other (please describe)

## Description

The device pages shipped as "coming soon" placeholders. This fills them in with real connectivity and a monitoring dashboard: whether a device is talking to the broker right now, how available it was over a window, how much data it sent and where it landed, what those payloads carried, and the raw measurement record behind the aggregates.

Connectivity comes from two independent sources, kept separate on purpose. Live presence is read from AWS IoT Fleet Indexing: a batched fleet-index search resolves each Thing's connected state and last-seen timestamp, and a device absent from the index reads as unknown rather than offline (the index may still be building). This becomes a `connectivity` field on the device list and detail responses, rendered as its own visual axis next to the credential-status badge and polled every 15s on the detail page. Historical presence comes through a new pipeline instead: AWS emits connect/disconnect lifecycle events, a dedicated topic rule archives them to S3, and bronze/silver/gold tables turn them into a deduplicated event log and a per-device last-activity table. The lifecycle events are deliberately routed to S3 only and never into the Kinesis measurement stream, because bronze consumes that stream unconditionally and cannot be reset.

The monitoring dashboard is warehouse-backed and range-scoped. A single use case resolves the device once, then runs five focused queries (sessions, throughput, battery, payload breakdown, recent measurements) in parallel and fails loudly if any one fails, so the dashboard shows a real error instead of a half-populated view. Last-data-arrival is served from a separate lightweight endpoint so the device list and detail never block on the SQL warehouse.

Turning the raw event log into sessions and uptime lives in one domain-pure transform, shared by the backend and its tests. It pairs connects and disconnects into sessions and handles the messy cases MQTT produces: repeated connects keep the earliest start, a disconnect from an older interleaved session does not close the live one, a disconnect leading the window implies a session already running at range start, and a session still open at the end runs to now. Events are capped at 1000 per window; past the cap the sessions themselves are still shown but uptime is withheld, since a percentage computed over a truncated window would read as fact while describing only the part that fit.

The dashboard UI is ordered the way an operator reads a device (healthy now, available over the window, data flowing, where it landed, what it carried, then the raw rows) with every time-series panel sharing the one selected range so a gap in one lines up with the gap above it. Experiment, protocol, and workbook ids are resolved to names only when the viewer has access to them, since a device publishing into an experiment says nothing about who may read it.

On the infrastructure side, the ECS task role gains `iot:SearchIndex` for the live lookups and the deploy role gains the fleet-indexing configuration permissions. Fleet indexing itself (an account and region singleton) and the Databricks read policy for the lifecycle-events S3 prefix are both behind opt-in variables, and a new `DEVICE_LIFECYCLE_EVENTS_S3_PATH` is wired through dev and prod.

## OJD-1602


apps/backend/src/iot/presentation/iot-device.controller.ts
```diff
@@ -116,6 +120,47 @@ export class IotDeviceController {
...
+
+      if (result.isSuccess()) {
+        return result.value;
+      }
+
+      return throwOrpcFailure(result, this.logger, "getIotDeviceActivity");
+    });
+  }
+
+  @CanAccess({ resource: "device", action: "read", param: "deviceId" })
```
dominikvrbic, line 142: Organization-boundary question rather than a definite blocker: I don’t think returning the raw device values is necessarily a problem when the viewer owns or can read the device, even if that device is used in an experiment they cannot access. The unresolved part is how much linked experiment/workbook/macro identity should cross that boundary: this route is authorized only by `device/read`, while the response can include experiment, protocol, workbook and macro IDs alongside GPS and samples. This is partly something the broader organization/resource work needs to settle. Could we document the intended ownership contract here and decide whether inaccessible linked-resource IDs should remain opaque server-side until that org model is in place?

apps/data/src/pipelines/centrum/silver/clean_device_lifecycle_events.py
```diff
@@ -0,0 +1,53 @@
...
+        dlt.read_stream(RAW_DEVICE_LIFECYCLE_EVENTS_TABLE)
+        .select(
+            F.col("clientId").alias("client_id"),
+            F.col("eventType").alias("event_type"),
+            (F.col("timestamp") / 1000).cast("timestamp").alias("event_timestamp"),
+            F.col("sessionIdentifier").alias("session_identifier"),
+            F.col("disconnectReason").alias("disconnect_reason"),
+            F.col("ingestion_timestamp"),
+        )
+        .withWatermark("event_timestamp", "7 days")
```
dominikvrbic, line 52: Could we revisit this seven-day event-time watermark? Once newer events advance the watermark, an older lifecycle event discovered later from an Auto Loader backfill—or delivered after more than seven days of pipeline downtime/out-of-order delay—can be dropped by the streaming deduplication. That conflicts with the comment above that the initial archive backlog and downtime must not lose history, and would permanently create incorrect sessions/uptime. Please use a deduplication strategy that preserves arbitrary historical backfill, or explicitly establish and test a bounded arrival-time guarantee; a regression test should ingest a newer event before one more than seven days older.

dominikvrbic (commented):
AI-assisted review disclosure: I commissioned two independent review agents for this PR: OpenAI Codex 5.6 Soul (high reasoning) and Kimi for Coding. The inline comments in this submission come from the completed Codex pass and were revalidated against current head `73f01a5`; the Kimi pass is still in progress, and I’ll add only any new, non-duplicate validated findings from it. One additional Codex finding about deriving sessions from the capped 1,000-event history is already covered by the existing unresolved CodeRabbit thread on `get-device-sessions.ts`, so I have not duplicated it.

apps/backend/src/iot/application/use-cases/get-iot-device-activity/get-iot-device-activity.ts
```diff
@@ -0,0 +1,51 @@
...
+      deviceResult.value.thingName,
+    );
+    if (activityResult.isFailure()) {
+      this.logger.warn({
+        msg: "Device last-activity lookup failed; rendering as unknown",
+        operation: "getIotDeviceActivity",
+        deviceId,
+        errorCode: activityResult.error.code,
+      });
+      return success({ lastDataAt: null });
```
dominikvrbic: This collapses two different states: “the device has never sent data” and “the warehouse lookup is unavailable.” The frontend treats `lastDataAt: null` as both “no data” and, for a connected device, `connectedButSilent`, so a Databricks outage becomes a device-health alarm and can send an operator toward the hardware. Could the contract preserve an explicit `unknown`/`unavailable` state (or a non-blocking activity error) and suppress the silent-device warning when the evidence source failed?

apps/backend/src/common/modules/databricks/databricks.adapter.ts
```diff
@@ -339,6 +349,352 @@ export class DatabricksAdapter implements ExperimentDatabricksPort {
...
+    thingName: string,
+    from: string,
+    to: string,
+  ): Promise<Result<DeviceFirmwareVersionRow[]>> {
+    const result = await this.runMonitoringQuery({
+      table: `${this.CATALOG_NAME}.${this.CENTRUM_SCHEMA_NAME}.clean_data`,
+      whereConditions: [["client_id", thingName]],
+      filters: [{ column: "timestamp", operator: "between", value: [from, to] }],
+      aggregation: {
+        groupBy: [{ column: "device_version" }],
```
dominikvrbic: Grouping the whole window by firmware version cannot produce transition history. For `1.0 → 1.1 → 1.0`, both `1.0` periods collapse into one row and the UI emits only `1.0 → 1.1`, omitting the rollback; repeated upgrades are lost similarly. Please derive chronological value changes (for example with `lag(device_version)` over timestamp), or rename the surface to a version mix/window and stop presenting it as an event log. A rollback case would make a useful regression test.

petar-INFO: You're right, grouping by version alone couldn't represent a rollback, the windows even overlapped. Changed the query to group per (time bucket, version) and the use case folds consecutive buckets into runs, so 1.0 → 1.1 → 1.0 comes out as three runs and the log shows both transitions. Rollback case is a regression test now. One caveat: transitions are bucket-granular (hour/day, same as the dashboard), so a flap inside a single bucket comes out with approximate ordering.

apps/backend/src/iot/application/use-cases/get-iot-device-activity/get-iot-device-activity.ts
petar-INFO: Agreed, this one was real. Added pipelineUnavailable to the activity response, so a failed warehouse lookup no longer pretends to be "no data ever". The tile shows "Unavailable" and the connected-but-silent warning doesn't fire when the lookup failed. Tests cover both states.

apps/data/src/pipelines/centrum/silver/clean_device_lifecycle_events.py
petar-INFO: Good catch on the comment, it promised more than the code guarantees, so I rewrote it. In practice the bound holds though: the broker rule writes events within seconds under date-partitioned keys, and Auto Loader lists those lexicographically, so backlog and downtime recovery always replay oldest-first and the watermark trails the data. The only way to lose rows is manually dropping old-keyed objects into the bucket after newer ones were processed, and for that a full refresh is the answer, not a wider watermark. Can't write the regression test you're asking for, we have no harness for DLT flows.

apps/backend/src/iot/presentation/iot-device.controller.ts
petar-INFO: The way I see it: everything in that response is data the device itself produced, so device read is the right gate. The linked experiment/workbook/macro ids go out as raw UUIDs and the client only resolves names against lists the viewer can already see, so a foreign id just shows up as "unknown" with no name or link. Whether we should strip foreign ids server-side is really a question for the org/resource-model work, I'd rather decide that once for all resource types than invent a rule here.

dominikvrbic:
## Related Linear tickets

This PR relates to the following Linear tickets (all assigned to Petar; each ticket has a "PR #1900" section with the landed scope and test notes):

| Ticket | Status | Relationship | Why |
|---|---|---|---|
| [OJD-1714](https://linear.app/openjii/issue/OJD-1714) — B8. Device connectivity, live fleet-index state, lifecycle-event history & monitoring dashboard | In Review | **Closes** | Dedicated ticket for this PR's exact scope; acceptance criteria and test steps mirror what the PR delivers. |
| [OJD-1602](https://linear.app/openjii/issue/OJD-1602) — (Epic) Devices & Sensor Calibration | In Review | **Related** | Fills in the Devices UI placeholder pages this epic owns, but calibration execution, reports, and the mobile surface stay open — an epic isn't closed by one PR. |
| [OJD-1609](https://linear.app/openjii/issue/OJD-1609) — B3. Web Devices page | In Review | **Related** | The pages shipped under this ticket as "coming soon" placeholders; this PR fills them with live connectivity and monitoring, but the calibration run action from its original scope is not part of this PR. |
| [OJD-554](https://linear.app/openjii/issue/OJD-554) — (Epic) IoT Data Resilience & Device Identity | In Review | **Related** | Extends the epic's S3 archiving approach to connect/disconnect lifecycle events (dedicated topic rule → S3 → medallion pipeline), but large-payload flows and certificate lifecycle work remain open. |
| [OJD-1462](https://linear.app/openjii/issue/OJD-1462) — Automate X.509 device certificate provisioning | Done | **Related** | Builds on the registry/identity delivered there (credential-status badge, per-device connectivity). Already done and running on prod — left closed. |

dominikvrbic (approved):

PR #1904, opened 2026-08-14
fix: correlate sequential workbook uploads

dominikvrbic:
## Summary

- require one stable `workbook_run_id` on every mobile workbook upload
- reuse it across sequential protocol cells and multi-device results; rotate only when the flow starts its next iteration
- preserve the run ID when a step is retried or revisited, and document that `(workbook_run_id, macro_id)` is not unique
- resolve repeated protocol and macro references from their exact workbook cells, so the mobile analysis summary displays the actual producer-cell position
- propagate `workbook_run_id` into `experiment_macro_data` and `enriched_experiment_macro_data`
- reserve both workbook ID columns against custom/question field collisions, including mixed-case variants

## Upgrade and data migration behavior

- mobile persistence moves to v2; any active flow saved by an older app version is deliberately discarded because it has no trustworthy run ID, returning the operator to experiment selection
- flows started on persistence v2 continue to pause and resume normally
- historical, imported, and legacy macro rows may remain `NULL` for `workbook_run_id`; this change does not backfill or re-run their macros
- the enriched materialized view recomputes once after the schema projection is deployed
- `workbook_run_id` is intentionally visible in the Processed Data explorer, consistent with Raw Data

## Validation

- mobile: 92 test files / 948 tests pass with coverage; TypeScript check passes
- API: 38 test files / 1,449 tests pass with coverage; TypeScript check passes
- focused final-review validation: 138 mobile tests and 42 API tests pass
- formatting and diff checks pass; dependency-aware web production build passes after merging current `main`
- independent final reviews completed with Kimi 3 and Opus; all actionable findings were addressed
- Centrum recovery update `3e743f4b-d6ab-4d28-acf9-0c0906679586` completed
- Bronze contains five v5 records under run `f2a302ff-eb4e-4acb-8769-4eebe9897b09` and five v6 records under run `6ed4cb5d-4337-4e61-847d-cd364860638a`
- joining macro rows to raw rows confirms both run IDs and five ordered outputs per run: `ground`/1 through `ambit_4`/5

## Live experiment context

The dev Ambit orientation workbook is published as v6 (`0862760a-8c5c-4bcf-a6f6-079320fb0ea5`). It uses one shared MultispeQ protocol and five dependency-free fixed-position macros, so physical position identity is explicit and does not depend on protocol-name inference.

The live macro table does not yet expose `workbook_run_id` directly because this PR is not deployed. Until deployment, the verified run ID is available by joining `experiment_macro_data.raw_id` to `experiment_raw_data.id`.

## Follow-ups

- [OPENJII-54](https://linear.app/openjii/issue/OPENJII-54/remove-or-wire-dead-measurement-flow-lifecycle-actions): remove or wire dead measurement-flow lifecycle actions
- [OPENJII-55](https://linear.app/openjii/issue/OPENJII-55/define-workbook-step-retry-and-resubmission-semantics): define retry and resubmission semantics
- [OPENJII-56](https://linear.app/openjii/issue/OPENJII-56/align-reserved-experiment-fields-with-the-gold-schema): align the full reserved-field inventory with Gold


dominikvrbic:
@petar-INFO Could you give this a quick review? The key behavior is one stable workbook_run_id across sequential mobile measurements, plus propagation into the macro Gold/enriched tables. The five-step Android run is verified end to end on the phone.

petar-INFO (approved):
Approving in substance. The mechanism holds up: `nextStep` rotates only on the iteration-wrap branch, which is the one place `nextStepState` returns a bumped `iterationCount`, and `analysis-node.tsx:222-251` awaits the upload before calling it. So the five positions share an ID by construction. Same story on the questions path, where `handleSubmitAndContinue` uploads before dismissing.

Pipeline side is a projection widening, not new plumbing, and per the Databricks docs adding a nullable column to a streaming table won't force a full refresh, so no sandbox macro re-execution. Two things worth putting in the description: historical macro rows keep NULL forever, and the enriched MV recomputes once (cheap, joins only).

Nothing below blocks.

**1. Half the new run-ID logic can't execute.** `startNewIteration`, `retryCurrentIteration`, and `finishFlow` have no callers anywhere outside the store and its tests, and the `setFlowNodes` mint at `use-measurement-flow-store.ts:117` is unreachable because the only production caller passes `[]`. Similarly, both `workbookRunId ?? uuidv4()` fallbacks in the upload hooks: each hook has exactly one caller, and both now always pass an ID, so the two "legacy caller" tests simulate something that doesn't exist. Making the param required would move that guarantee into the type system and delete both branches.

**2. `build-upload-payload` still says the opposite.** Line 72 writes the field conditionally and its test asserts it's omitted when unlinked, while the new comment in `use-measurement-upload.ts` says every batch always has one. Worth picking a side.

**3. "Rotate on retry" isn't what ships.** The reachable retry is `handleRetry` at `analysis-node.tsx:253`, which calls `previousStep()` and deliberately doesn't rotate. `retryCurrentIteration` does rotate but is dead. Either wire it up or drop the claim from the summary.

**4. The `setExperimentId` guard never fires.** At `handleStart`, `state.experimentId` is still undefined (that call is what sets it), so the equality is always false and it always mints, discarding the ID `setFlowGraph` minted during the selection preload. Harmless now, but the comment describes behaviour the code doesn't reach, and the one case where the guard *would* fire is reusing a run ID across two separate attempts.

**5. `clean_data.py:84-85` now contradicts itself:** "Null on single-device uploads". `CONTEXT.md` got updated, this didn't.

**6. `workbook_version_id` isn't in `RESERVED_EXPERIMENT_COLUMN_NAMES`,** one line from where you're adding `workbook_run_id`, even though gold raw emits it.

**7. Question:** retry walks back without rotating, so a user who finishes position 1, advances, then returns and resubmits gets two rows under one run ID for one position. Nothing downstream enforces uniqueness. If that's intended (one attempt, corrected value), a line in `CONTEXT.md` would keep the first person writing a query from treating `(workbook_run_id, macro)` as a key.

Not a finding, just a heads-up: `workbook_run_id` isn't in `MACRO_TABLE_CONFIG.exceptColumns`, so it shows up as a visible UUID column in the Processed Data explorer. Consistent with Raw Data, so probably intended, but it's a user-visible change that isn't in the description.

@dominikvrbic

PR #1906, opened 2026-08-17
feat: ojd 1638 rfc unified iam organization scoped access control phase 4

vladstoe:

- [x] Feature
- [x] Bug Fix
- [x] Refactor
- [x] Documentation Update
- [ ] Performance Improvement
- [x] Test Update
- [ ] Chore
- [ ] Other (please describe)

---

Fourth and final phase of the IAM rollout ([#1751](https://github.com/Jan-IngenHousz-Institute/open-jii/pull/1751) Phase 1, [#1790](https://github.com/Jan-IngenHousz-Institute/open-jii/pull/1790) Phase 2, [#1852](https://github.com/Jan-IngenHousz-Institute/open-jii/pull/1852) Phase 3). Phase 1 shipped the organization substrate, Phase 2 the authorization substrate, Phase 3 made grants the only source of resource access. All three were invisible: nothing in the product could create an organization, invite anyone to one, or make a team. This phase is the surface.

- **Organizations become a product.** Create one, give it a profile, invite people, manage roles, group them into teams, and list it in a public directory that outsiders can request to join.
- **Teams become reachable.** `can()` has evaluated team grants since Phase 2 — they were dead code because nothing could create a team. Teams are now a grantee alongside users and organizations.
- **Resources can move and organizations can end.** A resource transfers between organizations; an organization is deletable, but never while it still owns anything.

## Why

The authorization model has been organization-scoped since Phase 2 — every resource has an owning organization, and org role is the first thing `can()` evaluates. But the only organizations that existed were the personal workspaces provisioned at sign-up. The entire org tier of the access model was reachable only by a database write.

That left three things unresolved rather than merely unbuilt. A resource stranded in a dead personal workspace had no way out. A collaborators surface listed explicit grants while org-derived access — the tier that actually governs most reads — was invisible on it. And nothing could ever be cleaned up: an organization that owned resources could not be deleted, and there was no transfer to empty it.

## The model

**An organization owns resources; people reach them through it.** Three roles, fixed:

- **Owner** — everything, including settings, directory visibility, deletion, and granting the owner role.
- **Admin** — full control of the organization's resources plus the membership surface (invite, remove, change roles below owner, teams). Deliberately *not* settings.
- **Member** — reads everything the organization owns, and on experiments may contribute measurements and annotations.

There is no configurable base permission. The member role is fixed; `organizations.base_permission` remains in the schema with no readers.

**There is no active organization.** No switcher, no session state. Organization selection happens per action — a picker on each create form, a target picker on transfer.

**Personal workspaces are not organizations.** Every account gets one at sign-up, and it is excluded from the entire organization surface: never in the directory, no members, no invitations, no teams, no settings, never deletable, cannot be left. It is identified by a reserved `personal-` slug namespace, enforced server-side at create and rename — not merely hidden in the UI.

**Access has two sources and both are now visible.** A collaborators list shows synthesized rows for the owning organization's owners, admins and plain members alongside the explicit grant rows, with each person appearing exactly once and their access attributed to where it comes from.

**Precedence is highest-wins.** Where several paths grant access, the strongest applies. This is what `can()` has always done; the design docs described strict first-match and are corrected, not the code.

### Better Auth organization layer

The org plugin owns the writes — create/update, invite/cancel, add/remove member, role change, team CRUD. Everything below is configuration and hooks around it rather than a reimplementation.

- **Slug is a real format now.** BA validates only non-empty and unique. Organization slugs are lowercase alphanumeric with single interior hyphens, capped at the column width, and the `personal-` prefix is reserved. Without that guard, `create({slug: "personal-lab"})` mints an organization the whole phase misclassifies as personal — undeletable, memberless, permanently unmanageable.
- **One canonical role spelling.** BA validates a role *after* trimming but gates who may hand out the creator role on the *untrimmed* string, then stores it verbatim. So `" owner"` or `"member, owner"` passes its check and lands in the column; every reader trims, so such a row reads as a full owner that BA's own permission check and last-owner counters do not recognise. All three membership writes now refuse anything but one exact spelling.
- **Settings are owner-only for real.** BA's default `adminAc` carries `organization:["update"]`, so an admin could rename the organization. Stripped from our admin role.
- **Two read gates.** `get-full-organization` returned every invitation row — address, role, status, expiry — to any plain member, the same payload the list-invitations gate withholds. It also answered a missing organization and a non-member differently, so any signed-in caller could enumerate which ids and slugs exist. A before-hook now refuses non-members with one uniform answer (mirroring BA's slug-outranks-id precedence, so a slug probe behind a valid id is refused too), and an after-hook strips `invitations` for members below owner/admin using the same rule the list gate uses.
- **`visibility` is registered as an organization additional field.** BA silently strips unknown keys from its update body, so without this the directory toggle would 200 and change nothing.
- **The auto-created default team is off.** With `teams.enabled`, BA creates a team named after the organization on every create — a team nobody made, appearing in the teams list and the grantee picker.
- **Nothing accepts an organization invitation on your behalf.** Every admission is an invitation the invitee accepts explicitly, at any role — matching GitHub, which has no instant add at all. Signing in accepts nothing, and there is no direct-add endpoint; join-request approval and role promotions stay immediate because each already carries its own consent. `requireEmailVerificationOnInvitation` is set explicitly `false`; BA's heuristic silently enables it for our config shape, which would 403 accept and reject for unverified OAuth accounts.
- **The leave shield is hand-rolled.** `/organization/leave` fires no organization hooks at all, and the Nest adapter matches hook paths by exact string equality with no wildcards, so refusing to leave a personal workspace is its own path-matched before-hook.

### Nest organizations module (greenfield)

Every read, plus the join-request domain. Organization, invitation and team writes go through Better Auth rather than getting Nest endpoints — the one deliberate exception is the three join-request writes, whose approval inserts an `organization_members` row directly so the status flip and the admission are one transaction. Admission is centralized: that path and Better Auth invitation acceptance both go through a single transactional `admitMember`, which takes `FOR UPDATE` on the organization row and holds the shared membership cap that Better Auth's `membershipLimit` reads from the same constant. There is **no direct-add endpoint** — see the invitation-only note below.

- Directory (public, non-personal, searchable, paginated, with per-row membership status for CTA state), organization profile, and an access-scoped resource showcase — the four resource `findAll`s gain an `organizationId` filter, so each viewer automatically sees exactly what `can()` allows.
- Members list with profile joins. The org-wide **Outside Collaborators** view originally planned here was **cut during the org-pages redesign** (with its backend derivation); what shipped instead is per-resource collaborator attribution — each resource's collaborators tab states where every person's access comes from. The cross-resource inventory is tracked as a follow-up.
- `GET /users/me/organizations` for the pickers and nav, in place of the deferred `customSession` membership map — the session is cookie-cached for a day, so embedded memberships would go stale on every role change and bloat every request.
- Join requests, modelled on the existing experiment join-requests module: request and cancel against a public organization, list/approve/reject for owners and admins, approval inserting a plain member row in the same transaction as the status flip. Notification emails to owners and admins on a new request, and to the requester on a decision.
- Team reads with profile joins for both the teams UI and the sharing grantee picker.

### Collaborators state where access comes from

The surface only ever showed explicit grants, so an organization admin who already held full control looked like a stranger on it, and someone could "add" them at a tier that changed nothing.

- Beside the grant rows, the list now synthesizes individual rows for the owning organization's living owners and one summary row each for admins and for plain members holding no grant. Anyone with a grant is broken out into their own row and subtracted from those counts, so nobody is represented twice.
- A broken-out row states **effective** access and its source. An admin's stale Can view grant reads "Can edit · Redundant grant · Admin · via organization" with no tier control, since moving it would write a grant nobody's access reflects. A member's *raising* grant keeps its tier control and its badge.
- Revoking is labelled for what it removes and states the outcome: organization-derived access is not revocable from a resource page, and the confirmation says what the person keeps rather than hedging that other access "may" survive. Self-leave says it too.
- **The picker annotates instead of filtering.** It used to drop anyone holding a grant while offering anyone with organization-derived access as though a grant would do something. Every candidate now shows what they already hold; a raising tier stays selectable, an inert one is disabled with the reason, and someone no tier can help is shown unselectable rather than hidden — being hidden is what makes people ask why search is broken. Raise-vs-inert is computed from the shared access matrix, so the per-resource-type asymmetries come out right for free.
- Fed by a new resource-scoped `GET /{resourceType}/{id}/grantee-users` inheriting the controller's `can(share)` gate. Deliberately not an `organizationId` parameter on the global user search: that route also serves the member picker and the invite flows, and teaching it to answer "what is X's role in organization Y" opens a membership-disclosure question on a route with no resource scope.
- **One collaborator count across three surfaces.** `countCollaborators` counted owners plus grants and knew nothing about the summary rows, so it disagreed with the list beside it. It now computes what the surface renders, and the experiment overview's Details card reads the same function instead of counting contributors. The two are pinned against each other by a test comparing both outputs rather than by shared arithmetic — shared code cannot catch a divergence.
- **Organization members may contribute to their organization's experiments.** `roles.member.experiment` becomes read + contribute. The weakest *grant* already carried contribute on an experiment, so belonging to the owning organization conferred less than being handed the weakest share. Experiments only, and the public-passer-by case is untouched. Retroactive by construction: a matrix change, not a migration.
- `CollaboratorRow` became an exhaustive switch, which surfaced seven consumers reading `.grantee`/`.granteeId`/`.granteeType` off the union without narrowing. All compiled clean before; all would have mis-filtered or thrown on a summary row.

### Teams

Flat, organization-scoped, members only. Team CRUD and membership ride BA. In sharing, the grantee picker gains the owning organization's own teams, and `createGrant` validates that a team belongs to the resource's owning organization — a team from elsewhere would be access the owning organization cannot see, manage or account for. A team grantee is never an Outside Collaborator, and transfer keeps that true by dropping the source organization's team grants.

### Lifecycle

- **Resource transfer** for experiments, macros, protocols and workbooks (devices excluded — they have no transfer route, enforced by the type rather than a runtime refusal). The gate is two-part: `can(manage)` on the resource, **and** authority over the organization losing it — owner or admin there. A bare grant-holder can never pass, because grant roles carry `manage` and revoke is `share`-gated: any "Can edit" outside collaborator could otherwise move somebody else's work into their own workspace and lock the organization out of it.
- **The one exception is the case with nobody to expropriate from:** an organization where every owner *and* admin has closed their account. Its resources would otherwise be stranded, with nobody left to move them and no way to delete the organization while it owns them, so whoever still holds control through a grant may carry them out. Owners *and* admins, not owners alone — an organization that merely lost its owner is still running, and treating it as abandoned is exactly the expropriation this gate prevents.
- The transfer transaction re-asks both halves on its own handle once its locks are held, using the same `can()` evaluator rather than a second reading of the grant tables. A resource that moved since the decision was made is refused rather than moved again.
- **`can()` now returns the owning organization it resolved against.** A caller that authorizes and then re-reads the resource's organization has a window in which a transfer lands, leaving it acting on an organization nobody authorized for. One real instance is fixed by this: the grantee-team picker authorized against organization A and would return B's team names and member counts.
- **Organization deletion is blocked while the organization owns any resource** of the six owned types, and never cascades. Deliberate divergence from GitHub, which cascade-deletes repositories with the organization. Devices own live AWS Things and certificates that only their own delete path can tear down, and a research platform with monotonic "published stays published" visibility should not vaporize public resources behind one confirm dialog. The way out is to transfer or delete each resource. **Device groups**, which arrived with `main`'s device-group feature mid-review, join the same rule for a third reason: a group holds no work, but its grants are polymorphic and cleaned only by its own delete path, so cascading one away would leave access behind naming a group that no longer exists. They are counted in the blocker and named in a grant, but deliberately absent from the resources showcase and its totals — listing them means giving them a row, a label and a badge first.
- **Deleting an organization or a team tears down the grants naming it.** `resource_grants` is polymorphic on the grantee side, so it carries no foreign key and nothing cascades; the rows would otherwise survive as access nobody can see or revoke, and could be re-associated with a future organization or team reusing the id.
- **Account deletion also refuses while you are the sole living owner of a non-personal organization.** The pre-existing blocker is per-resource and escapes as soon as somebody else holds an admin grant, so an organization losing every living owner and admin was reachable. Checked pre-flight to drive the dialog and again inside the deletion transaction under the same owner-row locks the resource blocker takes. Personal workspaces are excluded and have to be — everyone is the sole owner of their own, permanently, so including them would refuse every deletion on the platform.

### Web surface

- New routes: organizations list and directory, create wizard, organization overview with a resource showcase, members (Members / Invited / Requests), teams and team detail, owner-only settings with a danger zone, and an **Invitations tab in account settings** beside General, Security and API keys.
- **Pending invitations reach you in the product, not only by email.** The header bell lists what is waiting, from Better Auth's own session-scoped `list-user-invitations`, and links to the account tab where each is accepted or declined. Rendered as its own section rather than as activity entries — that context is a per-tab, in-memory job tracker, and an invitation is neither a job nor client-derived. The indicator follows "something is waiting for you" rather than the unread counter, because an invitation stays actionable until it is answered, and a failed read shows as an error rather than as an empty list that would read as "nothing waiting". Expired rows are filtered out: Better Auth reports `status = 'pending'` and never retires an invitation that has lapsed, so unfiltered the bell would offer rows whose Accept can only fail.
- **Organization visibility is chosen at creation**, not defaulted on the user's behalf. The create wizard's Profile step carries the same two-radio-card control the settings surface uses, and Review reads the choice back with its consequence spelled out.
- **An organization picker on all five create forms**, defaulting to the personal workspace. One personal-workspace treatment across every picker, including transfer's target picker.
- A **Transfer** action on the four transferable types, gated on a new `canTransfer` capability mirroring the server gate — plain `canManage` is not enough.
- Failed reads no longer render as authoritative answers: a failed user search offers a retry rather than the invite-by-email path it exists to suppress; a failed deletion-blocker read keeps Delete disabled rather than enabling it on a defaulted count of zero; team grant counts render a skeleton while pending and an error state on failure, degrading per-card so a healthy list is never blanked.
- Every new string ships in en-US, de-DE and nl-NL.

### Rename release 2

Release 1 (Phase 3) taught every writer to emit `viewer` while readers still accepted `member`. This ships the other half: the three compatibility shims are reverted and the stored rows are rewritten. `normaliseTier` is kept as a totality guard, which is what makes a release-1 straggler harmless.

## Documentation

The user-facing half of the phase: a new **Access** section (`/guide/access` — the three routes access arrives by, and that the strongest wins), a new **Organizations** section, two Sharing pages for the owning-workspace picker and resource moves, and an **access-troubleshooting** page that starts from the symptom rather than the model. Eleven new pages, six corrected where Phase 4 changed the answer, seven screenshots captured through the real UI.

Eight `DocsHelpLink` CTAs now point at the page that answers them, gated by a new `check-platform-doc-links.mjs` — the existing checker only sees links *inside* the docs, so a platform CTA could 404 indefinitely.

Mermaid also renders now. Fumadocs ships the remark transform but no renderer, so fences had printed raw source on five developer pages ever since [#1815](https://github.com/Jan-IngenHousz-Institute/open-jii/pull/1815) dropped Docusaurus's mermaid theme. Lazy-loaded, so only the six diagram pages pay for it.

## End-to-end coverage

An end-to-end suite of **84 scenarios** drives the real stack against a real browser and database — all green. It lives in a **standalone repo**, not in this changeset, which is why it is not reviewable from the diff. The Phase 4 files are a disjoint set from Phase 3's with their own run script, so iterating on them never re-runs the older suite; the two coexist against one seeded database. Better Auth's own write paths — create, invite, accept, role change, team CRUD, delete — are exercised here rather than in backend specs, because the backend TestHarness mocks `@repo/auth/server` down to `getSession` and no protection hook executes under it.

| Spec | Area | n | Covers |
| --- | --- | --- | --- |
| **OC** | Create + directory | 11 | Slug guard, URL rule, private-unless-asked, publish-at-create, wizard people, My/All organization listings, membership scoping |
| **OI** | Invitations | 14 | Expiry, decline, wrong-account, accept-at-role, roster dedupe, no role auto-accepted at sign-up or at a later sign-in (member/admin/owner), registered-pick creates an invitation, admin cannot make an owner |
| **OM** | Members + roles | 7 | Last-owner refusals, admin-cannot-touch-owner both ways, owner-grants-owner, plain-member roster, removal ends access, personal-workspace refusals |
| **OF** | Transfer | 7 | Manage is not authority to move, target list contents, destination membership, grants kept vs source team grants dropped, archived still movable, all shareable types |
| **OP** | Create-form pickers | 7 | All four types land in the chosen organization, review states it first, untouched picker stays personal, no picker without memberships |
| **OJ** | Join requests | 6 | Request/withdraw/approve/reject, private organization neither discoverable nor joinable, no path for an existing member |
| **OT** | Teams | 6 | Create/open/rename, identity survives rename, member picker scope, plain-member read-only, deleting a team takes its grants |
| **OG** | Team grants | 5 | Teams source appears only where teams exist, team row is never an outside collaborator, tier fallback, revoke removes the access behind it |
| **OD**\* | Organization deletion | 5 | Block names every type owned including devices, empty-then-delete, grants held elsewhere go too, personal workspace has no danger zone |
| **OA** | Account deletion | 6 | Organization membership does not block, hand-off does not release a sole owner, living admin is not a husk, promote-to-proceed, personal workspace never blocks, leaving the platform leaves every shared organization too |
| **OX** | Overlapping access | 4 | Higher of direct vs team grant wins, admin authority not capped by a lower grant held, outside-collaborator badging, leaving keeps a direct grant |
| **OD**\* | Collaborator rows | 4 | Summary counts only people holding nothing of their own, raising grant keeps its control, redundant grant labelled and removable, picker refuses a no-op |
| **OV** | Overview showcase | 2 | Every owned type listed including devices, private rows badged and members-only |

\* **Two spec files share the `OD` prefix** — `o9-org-deletion` (OD1–OD5) and `od-collaborator-rows` (OD1–OD4). Harmless at runtime, since ids are only labels, but it makes "OD3" ambiguous in review and should be renumbered before anyone cites one.

<details>
<summary>All 84 scenarios</summary>

**OC — create + directory** · OC1 the slug is refused three separate ways before the wizard will advance · OC2 the website field applies the platform's URL rule · OC3 creating lands on the new organization, and it is private unless asked otherwise · OC3b an organization can be published as it is created, not only afterwards · OC7 a person collected in the wizard arrives at the role they were given · OC8 the settings form applies the same rule and stores what it is given · OC4 My organizations lists the shared ones as cards, never a personal workspace · OC9 a card's resource pill counts what the organization owns, across all five types · OC5 All organizations is what the caller may see: public ones and their own, never personal · OC10 the listing is membership-scoped, and a private organization is in neither filter · OC6 the sidebar reaches organizations, and a personal workspace has no page

**OI — invitations** · OI1 Invited holds the live invitations only, and an expired one blocks nothing · OI2 declining an invitation joins nothing · OI3 the wrong signed-in account is told so, and the invitation survives · OI4 accepting through the link joins at the invited role · OI5 somebody already on the roster is dropped from the results, and their address is explained · OI6 signing up with an invited address joins nothing until it is accepted · OI7 a plain member gets no management surface, and the endpoint behind it refuses · OI8 a registered pick creates an invitation rather than a membership · OI9 an admin cannot make an owner, in the dialog or behind it · OI10 signing up with an invited address does not take an Owner invitation · OI11 signing in again does not take an Owner invitation waiting for an existing account · OI12 signing up with an invited address does not take an Admin invitation · OI13m signing in again does not take a pending Member invitation · OI13a signing in again does not take a pending Admin invitation

**OM — members + roles** · OM1 the last owner has no role control, no remove, and a leave that says why · OM2 an admin cannot touch an owner in either direction · OM3 only an owner grants owner, and a second owner unlocks the first one's exit · OM4 a plain member gets the roster and their own exit, and nothing else · OM5 removing a member ends their access to a private organization · OM6 a member can leave, and a personal workspace refuses the same call · OM7 a personal workspace has no roster, for its own owner

**OF — transfer** · OF1 managing a resource is not authority to move it out of its organization · OF2 the target list excludes the current owner and includes the personal workspace · OF3 the caller has to belong to the destination · OF4 a transfer keeps user grants, drops the source's team grants and re-derives the owner · OF5 an archived experiment can still be moved out · OF6 a personal resource can be carried into an organization, and its page says so · OF7 every shareable type carries the same field and the same gate

**OP — create-form pickers** · OP1 an experiment created into an organization lands there, and the review says so first · OP2 a macro created into an organization lands there · OP3 a protocol created into an organization lands there, with the same review row · OP4 a workbook created into an organization lands there · OP5 leaving the picker alone keeps the resource personal · OP6 someone who belongs to no organization is shown no picker at all · OP7 the picker offers the shared organizations and the personal workspace, nothing else

**OJ — join requests** · OJ1 an outsider asks to join from the organization's page and it goes pending · OJ2 withdrawing puts the request back where it started · OJ3 an admin approves, and the requester is a member on the same render · OJ4 a rejected request leaves no membership and moves to the history · OJ5 a private organization is neither discoverable nor joinable · OJ6 a member is offered no way to ask for what they already have

**OT — teams** · OT1 an owner creates, opens and renames a team · OT1b renaming a seeded team changes its name and not its identity · OT2 the member picker offers the organization's members and nobody else · OT3 a plain member sees the teams and none of the controls · OT5 a team's page states what the team itself can reach · OT4 deleting a team takes the grants it held with it

**OG — team grants** · OG1 the Teams source appears only where the owning organization has teams · OG2 a team row is a Team with a head count, never an outside collaborator · OG3 membership contributes; the team grant is what adds managing · OG4 dropping the higher team falls back to the lower one, and dropping both ends it · OG5 revoking the team grant removes the row and the access behind it

**OD — organization deletion** · OD1 the block names every type still owned, devices included · OD2 emptying the organization enables the delete, and it goes through · OD3 the grants an organization and its teams held elsewhere go with it · OD4 a personal workspace has no danger zone and refuses the call behind it · OD5 a member sitting on the organization page when it is deleted is refused cleanly

**OA — account deletion** · OA1 belonging to an organization that owns things does not block leaving the platform · OA2 handing the resources off does not release a shared organization's sole owner · OA3 a dead owner with a living admin is not a husk, and the grant-holder may not expropriate · OA4 promoting a second owner is the way through the organization blocker · OA5 a personal workspace never blocks its owner from leaving · OA6 leaving the platform leaves every shared organization too

**OX — overlapping access** · OX1 a direct grant and a team grant at different tiers: the higher one decides · OX2 an organization admin's authority is not capped by a lower grant they hold · OX3 a grantee from outside the owning organization is badged as one · OX4 leaving the organization keeps a direct grant, which then reads as outside

**OD — collaborator rows** · OD1 a summary counts only the people who hold nothing of their own · OD2 a raising grant keeps its tier control, and says where else access comes from · OD3 a redundant grant is visible, labelled, and removable without a tier · OD4 the picker refuses a no-op and allows a raise

**OV — overview showcase** · OV1 the showcase lists every type the organization owns, devices included · OV2 a private row is badged as one, and only members see it at all

</details>

## Migration (`0047_iam_admission_and_owned_resource_locks.sql`)

One migration for the phase, per the one-per-PR convention, rebuilt in place each time `main` took the number ahead of it — `0043` through `0046` ride in from merged `main` PRs and are not ours.

**DDL**

1. `organizations.visibility` — `NOT NULL DEFAULT 'private'`, so new organizations are private and every existing one, personal included, lands private.
2. `organizations.slug` **becomes `NOT NULL`**, preceded by a defensive backfill minting `legacy-<id>` for any slug-less survivor of the 0035 member-guarded cleanup. The slug is the personal-workspace marker, and a nullable column gave "is this organization personal?" a third answer that two code paths already disagreed about. A no-op on every known database, and the only thing between that one row and a failed deploy.
3. `organization_join_requests`, mirroring `experiment_join_requests`, with a partial unique index on `(organization_id, user_id) WHERE status = 'pending'`.
4. **Every org-owned resource FK moves from `CASCADE` to `RESTRICT`** — experiments, protocols, macros, workbooks and device groups; `iot_devices` has been `RESTRICT` since `0039`. The pre-delete count runs before Better Auth's delete transaction opens, so a resource transferred in after it was destroyed with the organization. The constraint is what closes that window; the count survives to turn the common case into a refusal that names what is in the way.
5. Indexes on `(email, status)` for **both** invitation tables — the resource-invitation auto-accept lookup runs on every sign-in, the organization one serves every pending-invitation read — the recipient's own list behind the bell and the account tab, plus the Invited tab and the invite picker, and the legacy `invitations` table had no index at all — plus `organization_id` indexes on the four resource tables the new organization-scoped filters read.

**DML — rename release 2**

6. `resource_grants.role` and `invitations.role`: stored `member` → `viewer`, and the `invitations.role` column default moves to `viewer`.
7. `organization_members.role` and `organization_invitations.role` are **canonicalized** — whitespace- and comma-tolerant, in `owner > admin > member` precedence, touching only non-canonical rows so the sweep is re-runnable and a `NULL` invitation role is left alone. Legacy values like `"member, owner"` are invisible to Better Auth's last-owner guards, which exact-match `'owner'`.

Safe for two different reasons rather than one. `resource_grants` has no live writer of the old spelling, because release 1 already emits `viewer` there. `invitations.role` does still have one — release 1's invitation insert writes `member`, and those instances are up during a rolling deploy — but a row landing after this UPDATE is read correctly anyway, since the reader maps anything that is not `admin` to `viewer`.

**The vocabulary trap:** only those two columns carry a *grant* role. `organization_members.role` and `organization_invitations.role` carry Better Auth *organization* roles, where `member` is the correct current name — so no rename reaches them. Item 7 touches those columns for an unrelated reason and renames nothing.

**Precondition:** release 1 (the Phase 3 deploy) must be fully live on production before this deploys.


packages/auth/src/organization/lifecycle.ts
```diff
@@ -0,0 +1,136 @@
...
+ * four cascade, so they are destroyed with the organization. For a create that is
+ * tolerable — the resource was born in an organization on its way out. For a
+ * **transfer in** it is not: a resource that existed elsewhere is lost.
+ *
+ * Known and open, because there is no lock that closes it from here. Better Auth
+ * calls this hook *before* `adapter.deleteOrganization`, which opens its own
+ * transaction, and hands the hook no adapter or transaction handle — so a
+ * `FOR UPDATE` taken here commits and releases before the delete begins, and cannot
+ * order the count against a writer. Closing it needs the count to run inside Better
+ * Auth's delete transaction, which the plugin does not expose.
```
dominikvrbic: **Blocking — silent data loss.** This does not require Better Auth to expose its delete transaction in order to become safe. Experiments, protocols, macros, and workbooks still use `ON DELETE CASCADE`, so a resource transferred in after this count is destroyed with the organization. Please change every organization-owned resource FK to `RESTRICT`/`NO ACTION` (as devices already are), keep this preflight for the friendly error, and translate the constraint failure. Then the race becomes a refused delete instead of irreversible loss.

apps/backend/src/users/core/repositories/user.repository.ts
```diff
@@ -323,11 +351,11 @@ export class UserRepository {
...
-        // 2. Re-check the sole-admin invariant inside the transaction. The
-        //    pre-flight check in DeleteUserUseCase drives the hand-off UX but is
-        //    raceable on its own: two last admins deleting at once would both see
-        //    the other and both commit.
-        await this.assertNotSoleAdmin(tx, id);
+        // 2. Re-check the deletion blockers inside the transaction. The pre-flight
+        //    checks in DeleteUserUseCase drive the hand-off UX but are raceable on
+        //    their own: two last admins — or two co-owners of one organization —
+        //    deleting at once would both see the other and both commit.
+        await this.assertDeletionUnblocked(tx, id);
```
dominikvrbic, line 369: **Blocking — the check is correct, but the transaction leaves dead owner memberships behind.** OpenJII's blocker ignores profiles with `deleted_at`, while Better Auth 1.6.23's leave/remove/self-demote guards count raw owner rows. With owners A+B, A passes this check and deletes their account but retains the owner membership; B can later leave because Better Auth still sees two owners. The org then has no living owner or in-product recovery. After this check passes, remove the user from non-personal org/team memberships in the same transaction (retaining the personal workspace), or make every ownership-reducing path share a living-owner-aware lock/check.

packages/auth/src/organization/plugin.ts
```diff
@@ -0,0 +1,265 @@
...
+            // refuses the call — so there is nothing here to authorize.
+            //
+            // Both are read through explicit narrowings: the hook context is generic
+            // over every endpoint, and `activeOrganizationId` is an additional field
+            // this plugin contributes, so neither is typed at this point.
+            const query = ctx.query as { organizationId?: string } | undefined;
+            const activeOrganizationId = (
+              session.session as { activeOrganizationId?: string | null }
+            ).activeOrganizationId;
+            const organizationId = query?.organizationId ?? activeOrganizationId ?? null;
```
dominikvrbic: **Authorization bypass.** Better Auth resolves this endpoint with `ctx.query?.organizationId || session.session.activeOrganizationId`, while this hook uses `??`. A plain member with an active org can request `?organizationId=`: the hook keeps `""` and returns at line 205, then Better Auth falls back to the active org and serves the invitation list under its weaker member-only gate. Mirror Better Auth's `||` behavior or reject empty IDs, and pin this with a member-level regression test.

apps/backend/src/organizations/core/repositories/organization.repository.ts
```diff
@@ -0,0 +1,718 @@
...
+    role: string,
+  ): Promise<Result<AddOrganizationMemberOutcome>> {
+    return tryCatch(async () => {
+      const inserted = await this.database.execute<{ id: string }>(sql`
+        INSERT INTO ${organizationMembers} ("organization_id", "user_id", "role")
+        SELECT ${organizations.id}, ${userId}::uuid, ${role}
+        FROM ${organizations}
+        WHERE ${organizations.id} = ${organizationId}::uuid
+          AND ${isNotPersonalOrgSql()}
+        ON CONFLICT DO NOTHING
```
dominikvrbic: `membershipLimit: 100` is enforced by Better Auth but bypassed by this direct insert. Join-request approval and sign-in invitation auto-accept also insert directly without the cap, so the limit depends on the admission route and concurrent admissions can cross it. Please centralize these writers behind one transactional `admitMember` primitive that serializes the count, inserts idempotently, and retires competing pending invite/request state—or remove the configured cap if it is not a product invariant.

packages/database/drizzle/0044_opposite_toxin.sql
```diff
@@ -0,0 +1,42 @@
...
+-- Grant-role rename, second release. Safe for two different reasons rather than one:
+-- `resource_grants` has no live writer of the old spelling, because release 1 already
+-- emits `viewer` there. `invitations.role` does still have one — release 1's invitation
+-- insert writes `member`, and those instances are up during a rolling deploy — but a
+-- row that lands after this UPDATE is read correctly anyway, since the reader maps
+-- anything that is not `admin` to `viewer`.
+--
+-- Only these two columns carry a *grant* role. `organization_members.role` and
+-- `organization_invitations.role` carry Better Auth *organization* roles, where
+-- `member` is the correct, current name — they must not be rewritten here.
```
dominikvrbic: Leaving the current organization role `member` unchanged is correct, but this migration still needs to canonicalize legacy multi-role/whitespace values in `organization_members.role` and `organization_invitations.role`. Before the new hooks, the mounted Better Auth endpoints could store values such as `"member, owner"`; server readers choose the strongest token, the web strict-parses and falls back to `member`, and Better Auth's own owner counters do not consistently trim. Normalize both columns using the same precedence as `normalizeOrgRole`, and add migration fixtures for comma/whitespace variants.

packages/api/src/domains/organization/organization.contract.ts
```diff
@@ -0,0 +1,128 @@
...
+   * Resources showcase. Each type's access-scoped `findAll` does the filtering, so
+   * an outsider sees the public rows and a member sees everything they may read.
+   * `totals` counts the same access-scoped set the capped rows come from.
+   */
+  listOrganizationResources: oc
+    .route({ method: "GET", path: "/api/v1/organizations/{id}/resources", successStatus: 200 })
+    .input(zOrganizationIdPathParam)
+    .output(zOrganizationResources),
+  /** The roster. Members only: who belongs to an organization is not public. */
+  listOrganizationMembers: oc
```
dominikvrbic, line 71: The PR describes an organization-wide **Outside Collaborators** view derived across `resource_grants`, but the organization contract exposes only the member roster and the UI has only Members / Invited / Requests. Resource-level collaborator attribution does not provide the promised cross-resource inventory. Please implement the derived paginated endpoint/surface, or explicitly remove it from Phase 4 acceptance criteria and create a concrete follow-up ticket.

apps/backend/src/organizations/presentation/hooks/organization-auth.hook.ts
```diff
@@ -0,0 +1,111 @@
...
+  async handleOtpVerify(ctx: AuthHookContext) {
+    await this.acceptInvitations(ctx);
+  }
+
+  private async acceptInvitations(ctx: AuthHookContext) {
+    try {
+      const user = ctx.context.newSession?.user;
+      if (!user?.id || !user.email) return;
+
+      const result = await this.acceptInvitationsUseCase.execute(user.id, user.email);
```
dominikvrbic: **Product/security decision needed.** This auto-accepts every pending invitation with its stored role, including `owner`, without recipient consent. An inviter can make someone an owner; after the inviter deletes their account, the recipient can become sole owner of an unfamiliar org and be blocked from leaving or deleting their own account. Consider auto-accepting only `member` and requiring explicit acceptance for `admin`/`owner`, or otherwise provide consent/provenance and a safe escape path. Please cover the sole-owner scenario if unrestricted auto-accept is intentional.

dominikvrbic (commented):
Independent Phase 4 review synthesized from Kimi, Opus, and a final validation pass. Documentation-only findings are intentionally excluded while Vlad completes the docs.

I would keep this draft until the data-loss race, ownerless-organization path, invitation authorization bypass, admission-limit inconsistency, and legacy role cleanup are resolved. The organization-wide Outside Collaborators view is also described as Phase 4 behavior but is not implemented.

Two rollout decisions also need explicit sign-off: the directly dispatchable backend workflow can bypass migration 0044, and all existing plain organization members gain experiment `contribute` permission retroactively.

The inline comments contain the reproducible failure scenarios and focused fix directions.

dominikvrbic (approved):

vladstoe:
Thanks — all seven are closed in `cd4ecb0f7`.

**1 — a transferred-in resource could be cascade-deleted with the organization.** All
four FKs are `RESTRICT` now, as devices already were, so the race ends in a refused
delete. The preflight stays for the friendly breakdown, and a constraint failure that
does escape the window is translated into the same wording rather than a 500 — wrapped
around Better Auth's delete endpoint, since its dispatch turns anything that isn't an
`APIError` into a 500 before an outer layer can see it.

**2 — the blocker passed but the transaction left dead owner memberships behind.**
Fixed wider than owner rows: account deletion now sheds **every** organization
association in the same transaction — memberships at any role, team rows, pending join
requests, and pending invitations to the address about to be scrubbed — the treatment
grants already got. The personal workspace keeps its owner, as it must.

**3 — `??` let an empty `organizationId` skip the gate.** Mirrored Better Auth's `||`,
pinned with member-level regression tests. Same mismatch existed on the
`get-full-organization` gate; fixed there too.

**4 — the membership cap held on some doors and not others.** Centralized per your
first option: one transactional `admitMember` — `FOR UPDATE` on the organization row,
count under the lock, idempotent insert — with the limit as a shared constant Better
Auth's `membershipLimit` reads too. A refusal is an answer, not a failure: join-request
approval undoes its claim so the reviewer can retry once there's room. Since 7 deleted
two of the three writers, the cap on invitation acceptance is now Better Auth's own
check — present at `crud-invites.mjs:275-279`, before the status flip, same configured
limit.

**5 — legacy multi-role values survive the migration.** Canonicalized in the same
migration, both columns, your `normalizeOrgRole` precedence, whitespace-tolerant and
idempotent, with fixtures for the messy variants. The existing "must not be rewritten
here" comment was about the grant-role rename and stays true; the two rewrites are now
distinguished in the file.

**6 — the promised Outside Collaborators view doesn't exist.** Correct — it was cut
during the org-pages redesign, backend derivation included, and the PR text
over-promised it. Description fixed; what shipped is per-resource attribution, and the
org-wide inventory is tracked as a follow-up.

**7 — auto-accept granted `owner` without consent.** Your split is what we did first:
`member` was instant because it carries no responsibility, `admin` and `owner` waited
for an explicit accept. But promoting that member to admin or owner needs no approval
either, so the gap only moved a step later — and the same held for a registered pick,
added to the roster instantly at whatever role the actor chose. Simpler to close it
once, and it matches GitHub, which has no instant add at all: **every role now requires
the invitee to accept before they join.** So `addOrganizationMember` is deleted rather
than guarded, both callers invite, and the sign-in auto-accept path (use case, six
hooks, the repository serving them) is gone — Better Auth's accept endpoint, reached
from the accept page, is the only door in. Promotions and join-request approval stay
instant: the first comes after consent, the second *is* consent.

PR #1918, opened 2026-08-18
feat(backend): name export downloads after the experiment

### LINKED ISSUE #1643: Downloading files
The file names of the data when downloading are not helpful. It would be great to have at least something from the experiment e.g. the experiment name or so as the file name.

dominikvrbic:

- [x] Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Test Update
- [ ] Chore
- [ ] Other (please describe)

## Description

Downloaded experiment data arrived with names that said nothing about its origin:

```
clean_data-export-3f8a1c92-7b44-4e1f-9d02-ab7719e6f5c1.csv
```

Exports now download as `{experiment}_{table}_{timestamp}.{ext}`:

```
leaf-photosynthesis-2026_clean-data_20260102_030405.csv
```

Conventions are reused rather than invented. The slug rule matches the existing workbook download (`toLowerCase().replace(/[^a-z0-9]+/g, "-")`), and the timestamp reuses `compactTimestamp()` from the data-upload path, so `YYYYMMDD_HHMMSS` is already the house format for generated names.

Every segment degrades to an identifier rather than vanishing when metadata is thin: no usable experiment name falls back to the experiment id, an unknown table drops that segment, and a missing completion time falls back to the export id.

Timestamps are second-precision, so two exports of the same table finishing within the same second share a name and the browser suffixes the duplicate. That is a deliberate trade: the readable name is worth more than the uuid that guaranteed uniqueness.

## Related Issues

- Closes [OJD-1624](https://linear.app/openjii/issue/OJD-1624/downloading-files)
- Closes #1643

## Testing Instructions

- [x] `apps/backend`: `npx vitest run src/experiments/application/use-cases/experiment-data-exports src/common/utils/datetime.spec.ts src/common/modules/databricks/databricks.adapter.spec.ts src/experiments/core/repositories/experiment-data-exports.repository.spec.ts src/experiments/presentation/experiment-data-exports.controller.spec.ts` — 102 tests across 8 files
- [x] Same suite under `TZ=America/Los_Angeles`, to prove the timestamp does not drift with the server timezone
- [ ] Manual: open an experiment, export a table, download it from the export list, confirm the filename carries the experiment name

## Changes Made

- `export-filename.ts` (new): `buildExportFilename()`, with slugification, the fallback chain, and extension parsing
- `download-export.ts`: resolves the experiment to get its name, mirroring `list-exports.ts`
- `datetime.ts`: new `parseDatabricksTimestamp()`
- `databricks.adapter.ts`, `databricks.port.ts`, `experiment-data-exports.repository.ts`: thread `completedAt` through `streamExport`, read from the `completed_at` column the metadata query already returned

Three defects fixed on the way:

1. **Extension parsing.** The old `filePath.split(".").pop()` returns the *entire path* when the path has no dot, so a Spark `part-00000` file produced a filename with a path glued onto it. Now the basename is isolated first and the extension must match `\.[a-z0-9]{1,10}$`. Verified against the CSV, JSON, Parquet, and XLSX layouts written by `apps/data/src/tasks/data_export_task.py`.
2. **Timestamp drift.** Databricks SQL timestamps arrive with no offset, and `new Date("2026-01-02 03:04:05")` parses as *server-local* time. `parseDatabricksTimestamp()` forces UTC, matching what the repository already does for the API contract.
3. **Missing not-found check.** `download-export` never verified the experiment existed. It now 404s, consistent with the adjacent export use cases.

## Checklist

## Additional Notes

The generated name is pure ASCII alphanumerics, hyphens, underscores, and a single extension dot, so it is safe in the `Content-Disposition` header with no RFC 5987 encoding needed. Quotes, CR/LF, slashes, and traversal dots cannot survive slugification.

Confirmed the name actually reaches the browser rather than being a silent no-op: the controller returns a `File`, oRPC derives `Content-Disposition` from `File.name`, `main.ts` exposes that header via CORS, and `useDownloadExport` assigns `file.name` to the anchor's `download` attribute. No frontend change was needed.

Reviewed independently by a second agent before opening. It raised two comment-accuracy findings, both addressed here: the uniqueness comment now states the second-precision collision plainly, and NFKD alone left `ß`, `ø`, and `ł` to slug away to hyphens, so a small transliteration map now turns `Straße` into `strasse` instead of `stra-e`.

`mapCompletedRows` in the repository keeps its own timestamp normalization. It serves the list-exports *string* contract while the new helper returns a `Date`; consolidating them would mean an API-shape decision that does not belong in this PR.


petar-INFO (approved):

apps/backend/src/experiments/application/use-cases/experiment-data-exports/export-filename.ts
```diff
@@ -0,0 +1,64 @@
+import { compactTimestamp, parseDatabricksTimestamp } from "../../../../common/utils/datetime";
```
petar-INFO: this is not a properly written use case

petar-INFO (changes requested):

apps/backend/src/experiments/application/use-cases/experiment-data-exports/export-filename.ts
dominikvrbic: You're right, and thanks for catching it. It wasn't a use case at all, just a pure function, and it was the only non-`@Injectable` file under any `use-cases/` directory in the whole backend:

```
find src -path "*/use-cases/*" -name "*.ts" ! -name "*.spec.ts" -> only export-filename.ts lacked @Injectable
```

I checked how the codebase handles this elsewhere. Pure domain rules consumed by use cases live flat in `<domain>/core/` with a colocated spec:

- `visibility/core/visibility-transition.ts` (imported as `../../../core/visibility-transition`)
- `sharing/core/grantee-selectability.ts`
- `sharing/core/resource-staffing.ts`

And every `application/services/*.service.ts` is an `@Injectable()` class, so that wasn't the right home for a free function either.

So it's now `experiments/core/export-filename.ts` plus `export-filename.spec.ts` beside it, imported the same way the visibility use case does it. `use-cases/` is back to being only use-case classes.

Also fixed the overflow-date case CodeRabbit flagged in the same commit: `Date` rolls `2026-02-30` forward to March 2, so a malformed timestamp would have produced a plausible but wrong filename instead of falling back to the export id. `parseDatabricksTimestamp` now round-trips the parse and rejects it, with tests for that and for real leap days.

Full backend suite green on the rebased base: 231 files, 2685 tests.

petar-INFO (approved):

PR #1948, opened 2026-08-24
feat(mobile,iot): identify connected sensors and report stable device identity

dominikvrbic:
Linear: [OJD-1726](https://linear.app/openjii/issue/OJD-1726/device-lineage-physical-sensor-serial-resolution-sensormix-sensor-node)

Follow-up that makes these fields reach silver: [OJD-1757](https://linear.app/openjii/issue/OJD-1757/data-pipeline-surface-device-family-and-phone-provenance-in-silver) / #1994 (stacked on this branch).

## What

Mobile measurements now report **which sensor actually produced them** and a **stable identity** for both the sensor and the phone.

The original slice added `device_family` / `device_firmware`. Hardware testing against a real Ambit showed `device_family` was reporting `multispeq`, so this PR also fixes the cause.

## Why

`apps/mobile` never called `identifyDevice`. Every connection, Bluetooth and USB serial alike, was built as `new MultispeqDriver(...)`, and `MultispeqDriver` hardcodes `family = "multispeq"`. So `DeviceIdentity.family` was **always** `multispeq`, whatever was plugged in, and an Ambit capture uploaded `device_family: "multispeq"`.

It went unnoticed because the transports are compatible enough that the wrong driver still works: the MultispeQ driver carried the full Ambit protocol (`arrun,3,0,2,…`, 1750-char reply, checksum `7A1E3AA1`) without erroring. The app just had the wrong idea of what the device was.

The identification handshake already existed in `packages/iot` and has been used by `apps/web` (`useIotConnections`) since #1791. Mobile was simply never migrated.

## Changes

**Device identification**
- Mobile runs `identifyDevice` on connect and binds the driver the probe resolves. `multispeq-command-executor` → `identified-command-executor` (it is no longer MultispeQ-specific).
- Fixed the ambit hello sentinel. It was `/^(.*?)\s*\bready\b\s*$/im`, requiring "Ready" to end the line, but shipping firmware appends its build: `NEW AmbitV003 Ready FW:1.1.4-3-g2a76435-dirty`. Real devices fell through to `generic`; the `NEW Name Here Ready` fixture was the only reason CI looked green.
- `useBatteryLevel` reads `identity.batteryPercent` instead of sending the MultispeQ-only `battery` console command, which an Ambit rejects.
- `IDeviceDriver.cancel` is now optional: only MultispeQ has a device-side abort.

**Sensor serial**
- `AmbitDriver` lifts the trace's `sensor_id` onto `device_id`. That value is the ESP32 eFuse MAC, which the firmware formats specifically so traces "join ambit.device/1 on one stable identity" (`src/PAM.cpp:364`, `src/trace_v3.h:18`). Previously `device_id` was the Android USB device id (`1002`), which the code itself calls a weak fallback because it changes across replugs. A firmware-supplied `device_id` still wins.

**Phone identity**
- The install id derives from `Settings.Secure.ANDROID_ID` instead of a random `uuidv4`, so reinstalling no longer creates a new phone. Android's real serial is unreachable (`Build.getSerial()` has needed privileged `READ_PRIVILEGED_PHONE_STATE` since Android 10).
- Hashed into a deterministic `uuidv5`: deployed environments predate the relaxed serial contract below, and a raw SSAID fails their validation.
- This makes self-registration work as designed. `ensure-mobile-device` looks the phone up by `findBySerialNumber(installId)`; with a random id per install that lookup could never hit, so every reinstall created a new AWS IoT Thing and device row. Re-registering now returns the same `deviceId`.
- iOS keeps the random id: `identifierForVendor` resets once every vendor app is removed, so it is no more reinstall-stable than a uuid.

**Client metadata**
- Measurements carry `client_model`, `client_manufacturer`, `client_os`, `client_os_version`, `client_app_version`. Previously the phone model was sent only at registration and the OS version nowhere. `client_` matches the `client_id` the IoT rule already stamps, and is disjoint from the sensor's `device_*` namespace.

**API serial contract**
- `installId` was `z.string().uuid()`, but `serial_number` is a `text` column and the generic register contract already accepts MAC-shaped values ("Physical device identifier, e.g. MAC address"). Only the mobile endpoint had narrowed it. Both now share `zDeviceSerialNumber`. Strict widening: uuids still validate.

## Testing

Verified end to end on a physical Ambit over USB-OTG (CH9102 bridge) attached to an Android phone, running this branch through Metro against dev.

Same hardware, same experiment, before and after:

| Field | Before | After |
| --- | --- | --- |
| `device_family` | `multispeq` | `ambit` |
| `device_id` | `1002` | `10:91:A8:4F:53:48` |
| `device_firmware` | `1.1.4-3-g2a76435-dirty` | unchanged |
| `client_*` | absent | `NX789J` / `nubia` / `Android` / `16` / `1.1.0` |

Decoded from the persisted upload envelope (`measurements.db`, gzip+base64), and the measurement published to dev over MQTT with a `puback`. Device logs went from `[multispeq] tx(command=battery)` to `[ambit] identified family=ambit`.

Automated: mobile 1118 tests / 106 files, `@repo/iot` 348 / 22, `@repo/api` 1701 / 43. Mobile typecheck and ESLint clean.

## Notes for reviewers

- `AmbitDriver` validates commands against its own list, so a cell holding non-Ambit syntax now fails with "Ambit rejected the command: …" instead of being forwarded and returning the device's own `BAD COMMAND` text.
- Every device adopts the new install id on first launch, which changes existing thing names and forces one re-registration. Old uuid-keyed device rows become orphans that no phone will claim, so they want a cleanup pass.
- SSAID is scoped per app signing key, so a dev build and a store build register as two devices on one phone.
- **These fields do not reach silver yet.** Bronze parses with a fixed `StructType` (`raw_data.py:50` / `schemas.py:59`), and `device_family` is not in it, so it is dropped before silver. Bronze retains the raw JSON, so nothing is lost. A stacked PR adds the extraction using the `get_json_object` pattern already used for `client_id`, which avoids evolving the non-resettable bronze struct.

### DIFF
--- a/apps/backend/src/iot/presentation/iot-device.controller.spec.ts
+++ b/apps/backend/src/iot/presentation/iot-device.controller.spec.ts
@@ -301,11 +301,32 @@ describe("IotDeviceController", () => {
       expect(second.body.id).toBe(first.body.id);
     });
 
-    it("rejects a non-uuid install id (400)", async () => {
+    // installId is the device's serial, not a minted uuid: a phone reports a
+    // hardware id. It is bounded by length and the AWS IoT thing-attribute
+    // charset, which is what CreateThing would otherwise 500 on.
+    it("accepts a hardware-shaped install id (200)", async () => {
+      const response: SuperTestResponse<IotDevice> = await testApp
+        .post(testApp.resolveOrpcPath(contract.iot.ensureMobileDevice))
+        .withAuth(userId)
+        .send({ installId: "9774d56d682e549c" })
+        .expect(StatusCodes.OK);
+
+      expect(response.body.serialNumber).toBe("9774d56d682e549c");
+    });
+
+    it("rejects an install id outside the thing-attribute charset (400)", async () => {
       await testApp
         .post(testApp.resolveOrpcPath(contract.iot.ensureMobileDevice))
         .withAuth(userId)
-        .send({ installId: "not-a-uuid" })
+        .send({ installId: "not a serial!" })
+        .expect(StatusCodes.BAD_REQUEST);
+    });
+
+    it("rejects an empty install id (400)", async () => {
+      await testApp
+        .post(testApp.resolveOrpcPath(contract.iot.ensureMobileDevice))
+        .withAuth(userId)
+        .send({ installId: "" })
         .expect(StatusCodes.BAD_REQUEST);
     });
 
@@ -368,7 +389,7 @@ describe("IotDeviceController", () => {
 
       expect(response.body.id).toBe(device.id);
       // The owner of the device's org holds every action through that role, and no
-      // grant of their own — so there is nothing for them to leave. `canTransfer`
+      // grant of their own, so there is nothing for them to leave. `canTransfer`
       // is false even for them: a device's AWS Thing and certificate are
       // provisioned against its organization, so there is no transfer route.
       expect(response.body.capabilities).toEqual({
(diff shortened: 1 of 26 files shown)


petar-INFO (approved):
Approving. One reminder to track: `device_family` doesn't land anywhere downstream yet. Bronze keeps the raw `data` column, so nothing is lost in the meantime, and per the `client_id` precedent in `raw_data.py` the eventual landing is a `get_json_object` extraction rather than evolving the non-resettable `parsed_data` struct. Let's file a ticket so the field's purpose (distinguishing physical sensor families server-side) doesn't stay aspirational.

PR #1967, opened 2026-08-25
feat(backend,api): scope-related, paginated listings & tier-ranked search (OJD-1728)

dominikvrbic:
Backend half of the overview pages & search revamp (epic OJD-1727). Replaces the inconsistent legacy `filter=my|member` narrowing with a unified `scope=related` mode, adds paginated sibling procedures, relationship-tier ranking, and real cross-type scoring in global search. Search stays lexical — semantic/vector search is deliberately deferred (OJD-1730).

## What changed

**Contracts (`packages/api`)**
- New `shared/listing.ts`: `zResourceScope`, pagination query/envelope schemas, and `resolveListScope` (single home for alias precedence).
- All four list endpoints (experiments/protocols/macros/workbooks) gain `scope: "related" | "all"`; legacy `filter=my|member` kept as deprecated aliases mapping to `scope=related` (`scope` wins if both sent). For protocols/macros/workbooks this deliberately upgrades "my" semantics: shared and org-inherited resources now count — the OJD-1724 fix extended to the other three entities.
- Sibling `list*Paginated` procedures (`/api/v1/<entity>/paginated`) returning `{ items, page, pageSize, totalPages, totalCount }`. Existing array procedures' inputs/outputs untouched — deployed mobile keeps working unchanged.

**Backend (`apps/backend`)**
- `resourceTierExpression()` in `resource-access-scope.ts`: highest-wins CASE — owned 3 → shared 2 (user/team grant) → org 1 (org grant / owning-org membership) → public 0. Access control is unchanged and still gates everything before ranking sees it.
- Browse (no search term): `ORDER BY tier DESC` + existing per-entity secondary + `id` tiebreaker. Search: single weighted score = lexical rank + `TIER_WEIGHT (0.15) × tier`; cross-table bonus caps unified (`LEAST(sum, 0.10)`) so scores are comparable across entity types. Scoring is search-path-only (browse pays no scoring overhead).
- Global search: overfetches up to `limit` per type (was hard-capped at 8) and merges on the real score with `id` tiebreak — the 9th-best experiment can now beat every macro. Response contract unchanged.
- Separate `COUNT(*)` query for totals (correct on out-of-range pages); fixed a pre-existing missing tiebreaker in experiments search ordering; fixed `/paginated` route shadowing by `/{id}` on macros/workbooks (with per-entity route-resolution guard tests).

## Verification
- Backend suite: 266 files / 3168 tests green; `@repo/api`: 1699 tests green; monorepo `check-types` green (no web/mobile changes needed); lint clean.
- 25/25 live HTTP contract checks through the real OTP login (alias precedence, envelope shape, out-of-range totals, scope semantics, global-search shape).
- Adversarial review by a second agent: no critical findings (access gating on rows/COUNT/overfetch, injection surface, route order all confirmed); all fixable findings addressed in the second commit.
- Synthetic 40k-row EXPLAIN ANALYZE pass: tier ranking costs +50–100ms vs old ordering; all tier/access predicates index-driven (57k index scans on `resource_grants`, ~0 seq scans); deep pages cost the same as page 1. Browse-path scoring removal measured at 4.3× on workbooks.
- Validated end-to-end from a physical Android device against this branch: org-inherited private experiment appears via `scope=related` (invisible under the old filter), measurement flow works, offline precache intact.

## Known follow-ups (documented on OJD-1728)
- `TIER_WEIGHT=0.15` needs real-corpus rank-gap calibration (synthetic corpus has near-zero rank variance; an intent-pinning spec guards gross violations).
- Workbook global search ~2.2s at 40k scale — inherent to the score definition (4 access-scoped linked-entity EXISTS), not query shape (CTE rewrite measured at only ~6%); candidate fix is two-phase ranking, gated on real data.
- External API callers sending `filter=my` get the upgraded (wider) semantics — deliberate, this is the access-model fix.
- Legacy array procedures + `filter` aliases removed after web/mobile migrate (OJD-1732).

Part of OJD-1727. Implements OJD-1728.

### DIFF
--- a/apps/backend/src/common/utils/fts.ts
+++ b/apps/backend/src/common/utils/fts.ts
@@ -85,7 +85,10 @@ export const CROSS_TABLE_BONUS_CAP = 0.1;
 /**
  * Capped bonus for matches on related tables (creator, members, linked entities). One shared
  * ceiling across every entity keeps scores comparable in cross-type global search, where entities
- * differ in how many related tables they can match.
+ * differ in how many related tables they can match. The ceiling bounds the skew rather than
+ * erasing it: protocols and macros have a single probe so they top out at one step (0.05), while
+ * experiments and workbooks can reach the full cap. Bounded was preferred over identical because
+ * a flat any-match bonus would flatten within-entity ordering.
  */
 export function crossTableBonus(...matches: SQL[]): SQL<number> {
   const terms = matches.map(
--- a/apps/backend/src/common/utils/fts.ts
+++ b/apps/backend/src/common/utils/fts.ts
@@ -74,3 +74,27 @@ export function ftsMatch(vector: SQLWrapper, name: SQLWrapper, raw: string): SQL
 export function ftsRank(vector: SQLWrapper, name: SQLWrapper, raw: string): SQL<number> {
   return sql<number>`(ts_rank(${vector}, ${tsQuery(raw)}) + 0.3 * similarity(${name}, ${raw}))`;
 }
+
+/** Score added per relationship tier. Relevance dominates; tier breaks near-ties. */
+export const TIER_WEIGHT = 0.15;
+
+/** Per-match increment and the uniform ceiling on the summed cross-table bonus. */
+const CROSS_TABLE_BONUS_STEP = 0.05;
+export const CROSS_TABLE_BONUS_CAP = 0.1;
+
+/**
+ * Capped bonus for matches on related tables (creator, members, linked entities). One shared
+ * ceiling across every entity keeps scores comparable in cross-type global search, where entities
+ * differ in how many related tables they can match.
+ */
+export function crossTableBonus(...matches: SQL[]): SQL<number> {
+  const terms = matches.map(
+    (match) => sql`(CASE WHEN ${match} THEN ${sql.raw(String(CROSS_TABLE_BONUS_STEP))} ELSE 0 END)`,
+  );
+  return sql<number>`LEAST(${sql.join(terms, sql` + `)}, ${sql.raw(String(CROSS_TABLE_BONUS_CAP))})`;
+}
+
+/** Lexical relevance plus the tier bonus: the single key both browse and search order by. */
+export function searchScore(rank: SQL<number>, tier: SQL<number>): SQL<number> {
+  return sql<number>`(${rank} + ${sql.raw(String(TIER_WEIGHT))} * ${tier})`;
+}
--- a/apps/backend/src/common/utils/fts.ts
+++ b/apps/backend/src/common/utils/fts.ts
@@ -91,6 +91,12 @@ export const CROSS_TABLE_BONUS_CAP = 0.1;
  * a flat any-match bonus would flatten within-entity ordering.
  */
 export function crossTableBonus(...matches: SQL[]): SQL<number> {
+  // No probes means no bonus. Falling through would join an empty term list into
+  // `LEAST(, 0.1)`, which Postgres rejects at parse time.
+  if (matches.length === 0) {
+    return sql<number>`0::numeric`;
+  }
+
   const terms = matches.map(
     (match) => sql`(CASE WHEN ${match} THEN ${sql.raw(String(CROSS_TABLE_BONUS_STEP))} ELSE 0 END)`,
   );
--- a/apps/backend/src/macros/core/repositories/macro.repository.spec.ts
+++ b/apps/backend/src/macros/core/repositories/macro.repository.spec.ts
@@ -171,7 +171,7 @@ describe("MacroRepository", () => {
         .where(eq(macrosTable.id, macro.id));
 
       // They can still see it while they belong to the organization...
-      const before = await repository.findAll({ filter: "my", userId: author });
+      const before = await repository.findAll({ scope: "related", userId: author });
       assertSuccess(before);
       expect(before.value.map((m) => m.id)).toContain(macro.id);
 
@@ -182,7 +182,7 @@ describe("MacroRepository", () => {
           and(eq(organizationMembers.organizationId, org), eq(organizationMembers.userId, author)),
         );
 
-      const after = await repository.findAll({ filter: "my", userId: author });
+      const after = await repository.findAll({ scope: "related", userId: author });
       assertSuccess(after);
       // Authorship is not an access path: "My macros" must narrow what the caller
       // can already read, never hand back a body `can(read)` would refuse.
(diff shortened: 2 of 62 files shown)


apps/backend/src/common/utils/fts.ts
```diff
@@ -74,3 +74,30 @@ export function ftsMatch(vector: SQLWrapper, name: SQLWrapper, raw: string): SQL
...
+ * erasing it: protocols and macros have a single probe so they top out at one step (0.05), while
+ * experiments and workbooks can reach the full cap. Bounded was preferred over identical because
+ * a flat any-match bonus would flatten within-entity ordering.
+ */
+export function crossTableBonus(...matches: SQL[]): SQL<number> {
+  const terms = matches.map(
+    (match) => sql`(CASE WHEN ${match} THEN ${sql.raw(String(CROSS_TABLE_BONUS_STEP))} ELSE 0 END)`,
+  );
+  return sql<number>`LEAST(${sql.join(terms, sql` + `)}, ${sql.raw(String(CROSS_TABLE_BONUS_CAP))})`;
+}
```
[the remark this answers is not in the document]
dominikvrbic: Fixed in ab8aa5f08. `crossTableBonus` now returns a plain `0::numeric` when called with no probes, instead of joining an empty term list into `LEAST(, 0.1)`. No caller passes zero probes today, so this is a guard on the helper rather than a live bug; `fts.spec.ts` now covers both the capped and the empty case.

apps/backend/src/macros/core/repositories/macro.repository.spec.ts
```diff
@@ -1560,7 +1560,7 @@ describe("MacroRepository — list access scoping", () => {
       visibility: "public",
     });
 
-    const result = await repository.findAll({ filter: "my", userId: owner });
+    const result = await repository.findAll({ scope: "related", userId: owner });
```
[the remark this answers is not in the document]
dominikvrbic: Fixed in ab8aa5f08. Renamed to "keeps scope=related as a relationship view, unaffected by visibility" so the name matches the `scope: "related"` it actually passes.

apps/backend/src/workbooks/core/repositories/workbook-listing.spec.ts
```diff
@@ -0,0 +1,149 @@
...
+      expect(result.value.totalCount).toBe(5);
+    });
+
+    it("counts the scoped set, not the whole table", async () => {
+      const result = await repository.findPage(1, 20, { scope: "related", userId: stranger });
+
+      assertSuccess(result);
+      expect(result.value.totalCount).toBe(0);
+    });
+  });
```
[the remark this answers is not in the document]
dominikvrbic: Fixed in ab8aa5f08. Added a `findPage over the search ordering` block: six identically-worded workbooks (so they tie on score) walked across three pages, asserting no row is dropped or repeated, that the ids come back in ascending order (the `id` tiebreak being the only thing making the walk stable), and that a repeated identical query returns the same page.

apps/backend/src/workbooks/presentation/workbook.controller.spec.ts
```diff
@@ -17,6 +18,14 @@ import type { WorkbookVersionDto } from "../core/models/workbook-version.model";
 import type { WorkbookDto } from "../core/models/workbook.model";
 import { WorkbookVersionRepository } from "../core/repositories/workbook-version.repository";
 
+interface WorkbookPage {
+  items: { id: string }[];
+  page: number;
+  pageSize: number;
+  totalPages: number;
+  totalCount: number;
+}
```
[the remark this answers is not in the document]
dominikvrbic: Fixed in ab8aa5f08. `WorkbookPage` is now `WorkbookPaginatedList`, the type inferred from the same schema the contract declares as its output, so envelope drift fails this spec at compile time.

packages/api/src/domains/workbook/workbook.contract.ts
```diff
@@ -15,13 +15,19 @@ import {
...
+  zWorkbookPaginatedList,
+  zWorkbookPaginatedQuery,
 } from "./workbook.schema";
 
 export const workbookContract = {
   listWorkbooks: oc
     .route({ method: "GET", path: "/api/v1/workbooks", successStatus: 200 })
     .input(zWorkbookFilterQuery)
     .output(zWorkbookList),
+  listWorkbooksPaginated: oc
```
petar-INFO: consider reusing listWorkbooks with query param (or smth along those lines)

packages/api/src/domains/experiment/experiment.contract.ts
```diff
@@ -21,6 +23,10 @@ export const experimentContract = {
     .route({ method: "GET", path: "/api/v1/experiments", successStatus: 200 })
     .input(zExperimentFilterQuery)
     .output(zExperimentList),
+  listExperimentsPaginated: oc
```
petar-INFO: -||-

petar-INFO (approved):

packages/api/src/domains/workbook/workbook.contract.ts
dominikvrbic: Done in 963882c4e. The four `list*Paginated` procedures are gone; `list*` now takes optional `page`/`pageSize` and returns either the bare array or the `{items,page,pageSize,totalPages,totalCount}` envelope, selected purely by whether `page` was sent.

Neither param has a schema default, so an unpaginated caller stays unpaginated (the server applies the default page size only after it has decided the request is paginated). That keeps deployed mobile, which sends no page, on arrays at runtime.

For consumers, `@repo/api/shared/listing` exports `isPaginatedList` to narrow the union and `listItems` to pull the rows out of either shape, so nobody has to re-derive the check.

packages/api/src/domains/experiment/experiment.contract.ts
dominikvrbic: Done in 963882c4e. The four `list*Paginated` procedures are gone; `list*` now takes optional `page`/`pageSize` and returns either the bare array or the `{items,page,pageSize,totalPages,totalCount}` envelope, selected purely by whether `page` was sent.

Neither param has a schema default, so an unpaginated caller stays unpaginated (the server applies the default page size only after it has decided the request is paginated). That keeps deployed mobile, which sends no page, on arrays at runtime.

For consumers, `@repo/api/shared/listing` exports `isPaginatedList` to narrow the union and `listItems` to pull the rows out of either shape, so nobody has to re-derive the check.

packages/api/src/domains/workbook/workbook.contract.ts
dominikvrbic: (SHA correction: the branch has been rebased since, so this now lands in **d8576ce79** rather than the 963882c4e I cited above. Content unchanged.)

packages/api/src/domains/experiment/experiment.contract.ts
dominikvrbic: (SHA correction: the branch has been rebased since, so this now lands in **d8576ce79** rather than the 963882c4e I cited above. Content unchanged.)

Issue #12, opened 2025-01-06
Storyboard Visualization


dominikvrbic:
- Create visual storyboards to illustrate the desired future platform state.
- Validate the storyboard with stakeholders.

Chris-INFONL:
Part of 'user journey' frame on MIro - will be updated and expanded after more in-depth user scenarios have been defined (milestones 3 and 4)

PR #1989, opened 2026-08-28
feat(metrics): slot-shaped public metrics, a typed caption pool, and a scoped activity endpoint

petar-INFO:

- [x] Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Test Update
- [ ] Chore
- [ ] Other (please describe)

## Description

This turns the metrics groundwork into the actual public surface, and adds a private counterpart. It comes in two parts: a second wave of warehouse rollups, and the API that serves them.

The pipeline gains the rollups the richer UI needs beyond the base totals: an hourly activity profile (24 local-hour bins, for a daily-rhythm view), fixed activity windows (24h / 7d / 30d), daily activity broken out per experiment, an experiment-contributors window, featured-parameter stats (a parameter's 30-day count and median), and a pool-facts table of scalar highlights (median field-session size, a device's longest unbroken run of days, peak simultaneity, time zones spanned). Daily activity also starts carrying measurement volume in bytes.

The public endpoint is reshaped into a slot-shaped payload: independent, individually nullable slots for a hero triple (total measurements, total volume, time zones spanned), liveness (last measurement and 24-hour count), a 30-day community summary, the daily activity series, the hourly bins, the family breakdown, a featured parameter, and a captions array, each empty or null until the pipeline's first refresh. The captions are the notable part: a discriminated union of typed facts (a streak, a pace, a session size, an endurance run, a simultaneity count, a milestone, and so on) rather than sentences. The API ships data, never copy; the frontend keeps one i18n template per kind and skips kinds it does not know, so a new pool metric never changes the endpoint's shape. The payload is treated as the public disclosure boundary and holds exactly what the public UI renders.

Alongside it is a new authenticated scoped endpoint for an organization's or a user's own activity. It joins per-experiment warehouse rows against Postgres ownership and membership, aggregates to the requested scope, and returns the caller's activity next to the platform baseline and the caller's share of it; organization scope requires membership. The experiment- and user-grain inputs never leave the use case unaggregated, so the scoping is a real boundary, not a filter applied late. Both endpoints sit behind a dedicated cache adapter rather than an inline TTL, and both degrade the same way the groundwork established: a lagging or absent warehouse yields empty slots rather than an error, and a failed refresh serves the last snapshot stale so a warehouse blip never surfaces on the page.

### OJD-1756


dominikvrbic:
> [!NOTE]
> **This review was predominantly AI-generated.** Two independent agents (Kimi K3 and Codex gpt-5.6-sol at high reasoning) reviewed the PR head in separate read-only worktrees, and their findings were consolidated. Every claim marked **verified** below was independently re-checked against the code before posting; the rest is reported as the agents found it. Treat this as reviewer input to be confirmed, not as authoritative judgment, and please push back where it is wrong.

Two reviewers, run separately with different priorities (correctness plus the disclosure boundary; completeness plus failure modes). They converged on four issues and each found one the other missed. Kimi ran the new suites locally (35/35 pass); Codex could not run them (no local Postgres) but confirmed `pnpm --filter backend build` clean with 0 type errors and 803/803 contract tests passing.

---

## Blocking

**1. The "mine" scope misses every experiment a user joined.** (verified)

`getUserExperimentIds` unions `experiments.createdBy` with `experimentMembers` (`apps/backend/src/metrics/core/repositories/metrics.repository.ts:60-75`), but membership is now sole-sourced in `resource_grants`. Join-request approval inserts only into `resourceGrants` (`apps/backend/src/experiments/core/repositories/experiment-join-request.repository.ts:174-184`), and the only remaining non-spec references to `experimentMembers` in the backend are two cascade deletes (`experiment.repository.ts:567`, `user.repository.ts:375`). Nothing writes to it. So any user whose access came through the normal join flow sees an empty "mine" scope, and the endpoint effectively returns creator-only activity.

Suggested fix: query `resourceGrants` with `resourceType='experiment', granteeType='user', granteeId=userId`. The scoped spec only ever exercises the creator (`get-scoped-metrics.spec.ts:122-127`), which is why this passes today.

**2. Stale-on-refresh-failure is not implemented.** (verified)

The PR describes "a failed refresh serves the last snapshot stale so a warehouse blip never surfaces on the page", but `CacheAdapter.tryCache` (`apps/backend/src/common/modules/cache/cache.adapter.ts:28-51`) is a plain read-through: one TTL entry, no stale copy, no soft/hard TTL, no in-flight dedupe, no fetch-error fallback. At the 10-minute TTL (`metrics.module.ts:29-34`) the only snapshot is deleted on expiry, so there is nothing to serve stale.

Two consequences the reviewers highlighted:
- Public `load()` returns a non-null object even when every warehouse read failed (`get-public-metrics.ts:107-133`), so a fully-degraded all-nulls snapshot is written over the good one and served for a full TTL after a transient blip.
- Scoped `load()` returns `null`, which is never cached, so every authenticated request during an outage re-fires three warehouse queries including two full-table scans.

**3. The scoped endpoint errors where the public one degrades, and its schema cannot express "no data yet".** (verified)

Any scoped-daily, contributor, or windows failure (plus an empty `activity_windows` table) returns `null` from `load()` (`get-scoped-metrics.ts:85-99`), which `execute()` converts to an internal error (`:54-60`). The spec codifies this (`get-scoped-metrics.spec.ts:99-107`). This contradicts "both degrade the same way ... a lagging or absent warehouse yields empty slots rather than an error".

The root cause is in the contract, not just the code path: unlike the public slots, `scoped` and `baseline` are non-nullable (`packages/api/src/domains/metrics/metrics.schema.ts:98-105`), so first-refresh absence cannot be represented without inventing zeros.

**4. `hero.totalVolumeBytes` is not a total, and partial failures publish it as a measured zero.** (verified)

`buildHero` pairs an all-time `totalMeasurements` with a byte sum over only the fetched daily rows (`get-public-metrics.ts:136-151`). That fetch is `ORDER BY date DESC LIMIT 366` with no date predicate (`databricks.adapter.ts:1247-1265`), so it is "the newest 366 rows that have data", not a one-year window; since the pipeline emits sparse buckets it can silently span well over a year, and once there are more than 366 activity dates the hero under-reports volume while still calling it a total. `platform_totals` has no all-time byte column (`databricks.port.ts:3-12`).

Separately, `buildHero` gates only on `totals`: a failed daily read becomes `[]` and a failed pool-facts read becomes `null`, yielding `totalVolumeBytes: 0` and `timezonesSpanned: 0` (`get-public-metrics.ts:98-108,136-151`). That presents missing inputs as measured zero, which cuts against the rule applied deliberately to `community` two slots up ("hidden rather than shown with an invented zero"). Since `hero` is itself nullable, `hero: null` is the honest result unless all three inputs are present. Note commit 4b1241ad8 already fixed exactly this grain mismatch for `avgMeasurementSize`.

---

## Should fix

**5. One malformed warehouse row can 500 the public endpoint.** The metrics readers apply raw `Number(...)` / `String(...)` / `?? 0` with no row validation (`databricks.adapter.ts:1234-1244,1257-1265,1297-1306,1318-1324,1343-1348,1387-1408`). Missing columns become `0` or the literal string `"undefined"`; non-numeric values become `NaN`; an out-of-range hour reaches a schema capped at 23 (`metrics.schema.ts:28-31`). Controllers return these through the output-validated oRPC contract (`metrics.controller.ts:17-27`), so a single bad row produces a validation error rather than an empty slot, which is the exact inverse of the stated guarantee. The new adapter specs cover happy-path mapping only (`databricks.adapter.spec.ts:2617-2781`).

**6. Scoped and baseline figures labeled `30d` do not share a window.** The scoped filter is 30 inclusive UTC calendar dates anchored at request time (`databricks.adapter.ts:1372-1380`), summed as whole daily buckets. The baseline is a rolling 720-hour window anchored at pipeline-run time (`apps/data/src/pipelines/metrics/activity_windows.py:28-41`), and contributor pairs use that same rolling window (`experiment_contributors_window.py:34-43`). The oldest day contributes up to nearly 24 extra hours, and pipeline lag offsets the two windows further. The side-by-side comparison is the point of the feature, so this is worth getting exact.

**7. Scoped `lastMeasurementAt` is a date, not a timestamp.** It is set to the last daily bucket's `YYYY-MM-DD` (`get-scoped-metrics.ts:121-133`), while the identically-named field in the public `liveness` slot is a real instant (`metrics.schema.ts:10` vs `:94`). Same name, same contract, different grain. Consider `lastActivityDate`.

**8. Public timestamps pass through un-normalized.** `lastMeasurementAt` and `computedAt` are forwarded as raw Databricks strings (`databricks.adapter.ts:1241-1243,1304-1305`); the specs show offset-free values like `2026-08-28 10:00:00` (`databricks.adapter.spec.ts:2697-2723`), and the contract accepts any string (`metrics.schema.ts:9-12,79,94-105`). This adapter already has an ISO-normalizing helper at `:874-880`.

**9. Test gaps.** No test for a grant-based member in "mine" scope (the case broken in #1). No cache expiry, stale-fallback, or concurrent-miss assertions (`cache.adapter.spec.ts:164-200`). The public partial-failure test also fails `totals`, so it never exercises the daily-only or pool-only hero paths (`get-public-metrics.spec.ts:167-179`). No malformed-row cases, no rolling-vs-calendar boundary case, no coverage of the milestone-withheld branch (`get-public-metrics.ts:290-297`) or the streak recency cutoff (`:248`).

---

## Nits

- `contributorPairs` is left out of the failure-count logging (`get-public-metrics.ts:80-89`), so a contributor-pairs outage is invisible in logs while every other input is counted.
- Null becomes `0` in adapter mappings: a null `hour_local` lands in the midnight bin (`databricks.adapter.ts:1321`), same pattern for `median` (`:1347`).
- `scoped.measurements24h` is hardcoded `null` (`get-scoped-metrics.ts:133`), a permanently dead slot. Implement or drop it.
- No single-flight in `tryCache`: concurrent misses on one key fan out duplicate warehouse query sets.
- `getScopedDailyActivity` and `getContributorPairs` pull entire tables into memory (`databricks.adapter.ts:1372-1410`). Fine at current scale; worth a LIMIT or scope pushdown later.
- The 30-day window is defined in three places: both use cases and `apps/data/src/lib/openjii/openjii/metrics/constants.py:51`.
- Daily volume and count reductions are recomputed in both hero and caption assembly (`get-public-metrics.ts:145,213-214`); computing them once would keep the semantics from drifting.
- `DatabricksAdapter` is bound to `METRICS_DATABRICKS_PORT` via `useExisting` (`metrics.module.ts:24-27`) but only declares `implements ExperimentDatabricksPort` (`databricks.adapter.ts:57`), so port drift would not fail compilation. This matches the existing IoT pattern, so the consistency argument cuts both ways.
- Two new comments restate their symbol names rather than a constraint (`services/config/config.service.ts:133-136`, `databricks.port.ts:68-71`).
- Suspicion, unconfirmed: `countSharedExperiments` (`metrics.repository.ts:115-119`) treats any `granteeType != 'user'` grant as sharing, so granting an experiment to its own organization may count as shared.
- Suspicion: byte volumes are unbounded JSON numbers (`metrics.schema.ts:3-6,21-26`), losing precision above `MAX_SAFE_INTEGER`. Current scale is nowhere near it.
- The PR description mentions a 7d window and the caller's share of the baseline; neither appears in the code (`databricks.port.ts:31-39`, `metrics.schema.ts:89-106`). Probably just a description update.

---

## What holds up well

- **Authorization is genuinely solid.** Org membership is re-checked against Postgres on every request and before the cache is consulted (`get-scoped-metrics.ts:37-49`), so the controller is not the only boundary. `userId` comes only from the session, `scope` is a zod enum, `organizationId` is a validated uuid, and per-principal cache keys cannot be poisoned across callers. Both reviewers went looking for a crafted-scope bypass and neither found one.
- **The disclosure boundary is well designed.** Warehouse rows are filtered against Postgres-derived experiment IDs and aggregated before returning; neither reviewer found any path emitting experiment- or user-grain rows. Parameter names are allowlisted pipeline-side, medians are used instead of means, and no experiment names, user identifiers, or per-run data appear in either payload.
- **The typed caption union is the right call**, and the honest touches are the kind usually skipped: milestone withheld rather than dated at the window edge, streak requiring day adjacency, community slot hidden rather than zeroed. Issue #4 is worth fixing precisely because the rest of the file is this careful.
- Every declared public slot and all 11 caption variants have a producer, with no served-but-undeclared fields. OpenAPI matches the zod shapes. The new env var is wired consistently across `.env.example`, `.env.test`, config loading and validation, the devkit manifest, and dev/prod Terraform.
- Controller specs parse responses with the public zod schema, doubling as contract-conformance checks.

dominikvrbic:
## Multi-model review summary

> [!NOTE]
> **Disclaimer: this review was predominantly AI-generated.** Two models (Kimi K3 and Codex 5.6 Sol) reviewed this branch independently in separate read-only worktrees at `e04e8ec5f`. I spot-checked the load-bearing claims against the source before posting and have marked each finding with its verification status. Treat it as a prioritized list of things to look at, not as a verdict. Nothing here has been fixed or pushed.
>
> Method note, for calibration: Kimi reviewed open-ended. Codex's first pass came back off-brief, so its second pass was additionally asked to adjudicate a specific list of hypotheses. Where both models surfaced something independently, that is called out below, and those are the highest-confidence items.

Neither model found a data-exposure bug. Per-experiment and per-user warehouse rows are aggregated before response construction in both use cases, org scope re-checks Postgres membership on every request *before* the cache is consulted (so revocation is immediate), and no injection path exists: metrics table names are constants, values go through the escaping query builder, and scoped identifiers go through Drizzle parameters. Both models checked this specifically and both cleared it.

The substantive findings are about honesty of the numbers, cache behavior under failure, and unbounded warehouse reads.

### Both models flagged independently

**1. `hero` mixes denominators and invents zeros.** `get-public-metrics.ts:145-150` sums `volumeBytes` from at most 366 activity rows and returns it as `totalVolumeBytes` next to all-time `totals.totalMeasurements`. Because `daily_activity` is sparse by date, it is not even a fixed 366-day denominator. If the daily read fails while totals succeeds, `totalVolumeBytes` is `0`; if pool-facts fails, `timezonesSpanned` is `0`. Both are invented zeros inside an otherwise non-null hero, which is the exact thing the `community` slot comment says is avoided by hiding the slot instead. The `avgMeasurementSize` caption comment shows the window-vs-total trap was already on your radar; the hero has the same issue without the caveat. Note that `get-public-metrics.spec.ts:106-110` currently encodes the mixed denominator as expected behavior (`1_000_000` measurements against `6_000_000` bytes).

**2. Every scoped cache miss pulls the whole platform.** `getScopedDailyActivity` reads 30 days across all experiments with no `LIMIT` (`databricks.adapter.ts:1372`) and `getContributorPairs` reads the entire `experiment_contributors_window` table (`:1397`), then `get-scoped-metrics.ts:101-119` filters in JS down to a handful of rows. Cost is O(platform), not O(caller scope), and with per-user/per-org keys on a 10 minute TTL, N active users means N full platform reads per window. Pushing the id list into the query (`WHERE experiment_id IN (...)`, which leaks nothing since ids stay server-side) or caching these shared inputs under their own keys would both fix it.

**3. No miss coalescing.** `cache.adapter.ts:28-50` is a plain get / fetch / set with no single-flight guard, so a burst right after the public key expires makes every concurrent request run all nine warehouse reads plus three Postgres queries. This sits against the use case's own claim that anonymous page views "never fan out to the warehouse directly". The in-memory store is also per-replica, so this multiplies by instance count and snapshots differ between replicas.

### Kimi K3 only

**4. A degraded snapshot is cached for the full 10 minute TTL.** *(Verified.)* `load()` in `get-public-metrics.ts` always returns a non-null object, so when the warehouse is down `tryCache` happily stores the empty snapshot. A 30 second warehouse blip that lands on one cache miss serves an empty snapshot to everyone for 10 minutes after recovery. Consider not caching when the core reads all failed, or a much shorter TTL for degraded snapshots.

**5. Dead failure branch.** *(Verified, and the same root cause as #4.)* Because `load()` never returns `null`, the `"Public metrics are unavailable"` branch at `get-public-metrics.ts:50-52` is unreachable. Fixing #4 by returning `null` when everything failed would also make this branch live.

**6. Timestamps cross the public API as raw warehouse strings.** *(Verified: `toIsoOrNull` exists at `databricks.adapter.ts:874` and is used at 12 call sites for monitoring timestamps, but none of the metrics readers use it.)* Databricks `JSON_ARRAY` renders timestamps as timezone-less `yyyy-MM-dd HH:mm:ss`, which is exactly what the fixtures use. `liveness.lastMeasurementAt` and `computedAt` will be parsed as *local* time by `Date.parse` on the frontend despite being UTC, so they are off by the viewer's offset, and Safari may not parse the space-separated form at all.

**7. `milestone.ordinal` is not an ordinal.** *(Verified at `metrics.schema.ts:60`.)* It carries the milestone value (`1_000_000`), not a position. The next contract consumer will misread it.

**8. Stale comment.** `metrics.repository.ts:104-105` cites "the creator's own seeded control grant", but creation no longer seeds a creator grant (`create-experiment.spec.ts:104`, "seeds the creator with no grant at all"). The `or(ne(...), ne(...))` guard is still correct and still needed for rows from the org-backfill migration, which *did* grant the creator admin. The comment should point there.

**9. `readMetricsTable` duplicates `runMonitoringQuery`** (`databricks.adapter.ts:1198-1220` vs `:850-864`) build for build. Since `metricsTable()` already fully qualifies `catalog.schema.table`, the `schema` argument to `executeSqlQuery` does not affect resolution.

### Codex 5.6 Sol only

**10. Truncated results are never checked.** *(Verified.)* `sql.service.ts:216` returns `truncated` in `SchemaData`, and grepping `apps/backend/src` shows the only other references are in test fixtures. No production consumer reads it, and no metrics reader follows additional result chunks. Once either unbounded platform-wide query gets chunked, a caller whose experiments fall outside the inline chunk silently gets incomplete results and aggregates under-report rather than failing. This compounds #2, and Kimi independently noted the related risk that these unbounded reads rely on the Databricks 25 MiB inline disposition cap never being hit.

**11. `max: 500` may not be bounding anything.** *(Version premise verified: `@nestjs/cache-manager@^3.1.0` with `cache-manager@^7.2.8`. The runtime behavior I could not confirm, `node_modules` was not installed in the review worktrees, so please check this one directly.)* Codex's claim is that on this path Nest forwards `ttl`, `refreshThreshold`, and `nonBlocking` but drops `max`, leaving a Keyv-backed `Map` that only evicts an expired entry when that same key is read again. If so, one-off `scoped-user-<uuid>` keys accumulate unbounded. Worth a quick runtime check given scoped keys are per-user.

**12. Schema drift fails open, not closed.** String columns are coerced with `String(row[index.x])` at `databricks.adapter.ts:1258-1263`, `:1277-1282`, `:1343-1348`, `:1387-1408`. A missing column index yields the literal `"undefined"`; a null cell yields `"null"`. These still satisfy the contract's `z.string()` and reach the HTTP body, so a public activity date, family name, or parameter name can render as `"undefined"`. Numeric columns use `?? 0`, inventing plausible zeros. A missing user column would invent a contributor named `"undefined"`.

**13. Scoped and baseline windows are not comparable.** Scoped `measurements30d` covers 30 inclusive UTC calendar dates chosen at request time (`databricks.adapter.ts:1372-1380`), while `baseline.measurements30d` is a rolling 720 hour window computed at pipeline refresh (`apps/data/src/pipelines/metrics/activity_windows.py:28-44`). Measurements early on the oldest included date land in scoped but outside baseline. Pipeline lag worsens it around midnight, and since contributor pairs use the pipeline's window, one response can carry three different cutoffs while presenting them side by side.

**14. Scoped metrics 500 instead of degrading.** `get-scoped-metrics.ts:91-99` returns an internal error when the scoped warehouse tables are absent or lagging, so on first deploy the endpoint errors rather than returning empty slots. This contradicts the PR description's claim that both endpoints degrade gracefully. The contract has no nullable scoped slot, so either the contract or the behavior needs to move.

**15. `inArray` has no hard cap.** `metrics.repository.ts:30-42` receives a deduplicated list bounded by the count of active experiments in the 30 day contributor table. Not a practical concern today, but past 65,535 it exceeds the PostgreSQL bind-parameter limit and silently hides the whole community slot.

**16. `contributorPairs` is missing from the failure count.** `get-public-metrics.ts:80-89` omits it, so if it alone fails the community slot goes null with no warehouse warning. A Postgres failure inside `countInstitutions` is likewise converted to `null` without logging. The existing test asserts the hidden slot, not the missing observability.

### Explicitly refuted

Worth recording so these do not get re-raised:

- **SQL injection: refuted.** The caller controls no table or column name; escaping happens at `query-builder.base.ts:37-46`.
- **Caption typing merely asserted: refuted.** `MetricsCaption` is inferred from the Zod discriminated union, `buildCaptions` returns `MetricsCaption[]`, and the oRPC output contract validates at runtime. The typing is genuinely enforced.
- **Cache key collisions: refuted.** The `metrics:` and `macro:` prefixes separate domains and the `scoped-org-` / `scoped-user-` prefixes separate identities. The cache problems above are about retention, locality, stale behavior, and stampedes, not collisions.
- **Caption builders untested: refuted.** This one was my own framing error, not a model's. `get-public-metrics.spec.ts:122-145` does cover streak, pace, milestone, average size, and endurance.

### Test coverage gaps

Ranked roughly by risk:

- The timestamp passthrough (#6) is invisible to tests: fixtures already use the warehouse format, but nothing asserts the output format or ISO-ness.
- The `BETWEEN` window math at `databricks.adapter.ts:1372-1376`, the most off-by-one-prone code in the PR, is untested. The adapter spec covers row mapping only.
- Caching behavior for a degraded snapshot (#4) is unpinned: the "warehouse down" controller test runs once and never checks what a second request serves.
- Adapter tests (`databricks.adapter.spec.ts:2617-2782`) are happy-path row mapping plus one empty totals row. No coverage of missing or null columns (#12), per-reader SQL failure, truncated or multi-chunk results (#10), or the generated date bounds.
- Caption edges: a stale streak (newest row 2+ days old, forced to 0), a milestone crossing that predates the fetched window, threshold boundaries, and unsorted or zero rows.
- No HTTP test for a populated payload on either endpoint; `scoped-metrics.controller.spec.ts:62-73` mocks scoped rows to `[]`. No HTTP-level test for `scope=organization` without `organizationId` returning 400 (use-case level only).
- Cache tests skip TTL expiry, read/write errors, concurrent miss coalescing, the ignored `max`, and multi-instance behavior.

### Structural

Lower priority, included for completeness:

- `databricks.port.ts:68-71` and `cache.port.ts:11` have comments that only restate the declaration, against the AGENTS.md comment rule.
- `cache.adapter.ts:4-20` imports and explicitly implements both `MacroCachePort` and `MetricsCachePort`, so every future cache-using domain must edit shared infrastructure to add another structurally identical interface. A domain-neutral adapter bound per module would avoid that.
- `GetPublicMetricsUseCase` (`:57-299`) coordinates nine warehouse reads, runs Postgres counts, shapes the DTO, and owns every caption algorithm in one 301 line class, so caption-policy and snapshot-fetch changes touch the same file. Both models suggested extracting a caption builder.
- `scoped.measurements24h` is `z.number().nullable()` in the contract but the use case always sends `null` (`get-scoped-metrics.ts:133`). Harmless, but it is contract surface nothing produces.

---

Highest value items if you want a short list: **#4 and #5** (one fix covers both), **#6** (user-visible wrong timestamps), **#10** (silent under-reporting), and **#2** (the scaling shape this locks in).

petar-INFO:
@dominikvrbic thanks, applied

dominikvrbic (approved):

PR #1991, opened 2026-08-28
feat(metrics): surface platform activity end to end, from the public landing page to per-resource strips

petar-INFO:

- [x] Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Test Update
- [ ] Chore
- [ ] Other (please describe)

## Description

The public metrics API had landed, but nothing rendered it and the only activity numbers anywhere were the dashboard's static stat tiles. This turns metrics into a running sense of platform activity that shows up in three places: the public landing page, a signed-in user's dashboard and experiment overview, and a strip on every resource list row. It reaches across the stack, adding the read endpoints and the pipeline rollups those surfaces need, so despite starting as the presentation branch it is now a full-stack feature.

On the public landing page, a metrics section leads with a hero sentence and lays out the activity trend, the sensor-family breakdown, and two "most measured parameter" lines split by where the value comes from, a macro-derived one and a raw-instrument one, under a rotating pool of typed captions the frontend turns into copy through one template per kind. The charts render on the platform's existing plotly component rather than pulling in a second charting library, plotted on a real date axis with the chart chrome stripped back to a sparkline feel. The section degrades honestly: it drops entirely if the backend is unreachable, hides the instrument split while a family's publishers are still unresolved, and only shows captions whose underlying figure isn't degenerate.

Inside the app, the dashboard's stat tiles give way to a research-activity panel with a trend, the experiment overview gains an activity pulse, and every list page gets an activity summary above it and a per-row activity strip. Those strips are fed by a new resource-activity endpoint that is gated by each list's own visibility predicate, so a row only shows activity for a resource the viewer is already allowed to see, and an experiment-scoped endpoint sits behind the same read check the experiment pages use. The personal chips guard against the scoped slot being null before the first refresh.

The pipeline grows to match. It computes per-resource daily activity for protocols, workbook versions, and macros to back the strips, splits the featured parameters by provenance and ships their display labels from the pipeline rather than hardcoding them in the UI, widens the parameter window to 90 days, allowlists the Ambit vocabulary alongside MultispeQ, and states plainly when an instrument split is unresolved instead of guessing. A few facts are made honest and cheap at the source: pace is the true ratio, the arrival gap is a mean over the measured span since bulk ingest zeroes the median, and the streak, gap, and all-time volume are precomputed in the pipeline rather than derived per request.

## Linear issues 

Closes OJD-1756


dominikvrbic (approved):

dominikvrbic:
## AI-driven review

This comment was generated by AI from independent GPT-6 Astra, GPT-5.6 Sol, and Claude Opus reviews, then consolidated and checked against the source by the coordinating AI agent. Kimi K3 could not run because its provider rejected the request for insufficient credits. This is not a human sign-off.

Reviewed head `5c36fde5e3af91578f9d747fb8244bef40dc16d0` against base `76bc5fc1449b57f44110fd379b6b2071ae6a6df2`.

**Recommendation: address the following before merging, especially the two P1 findings.**

### P1

1. **Optional warehouse metrics block core pages.** [The experiment list](https://github.com/Jan-IngenHousz-Institute/open-jii/blob/5c36fde5e3af91578f9d747fb8244bef40dc16d0/apps/backend/src/experiments/application/use-cases/list-experiments/list-experiments.ts#L95) awaits warehouse activity after Postgres has returned its rows; protocol, macro and workbook lists do the same. [The homepage](https://github.com/Jan-IngenHousz-Institute/open-jii/blob/5c36fde5e3af91578f9d747fb8244bef40dc16d0/apps/web/app/[locale]/page.tsx#L82) also waits before returning its content. On cache misses, SQL permits a 50-second initial wait, a 60-second HTTP timeout and further polling. Uncached failures repeat the delay. A deferred warehouse stub reproduced the blocked list response. Use a short metrics deadline or background refresh; render the homepage section independently.

2. **Sensor extraction misses supported payload structures.** [parameter_stats.py](https://github.com/Jan-IngenHousz-Institute/open-jii/blob/5c36fde5e3af91578f9d747fb8244bef40dc16d0/apps/data/src/pipelines/metrics/parameter_stats.py#L95) reads `$[0].data.<name>`, but supported MiniPAR/PhotosynQ samples put scalars in `sample[].set[]`. Upload and silver processing preserve that structure. For example, the MiniPAR fixture has `345.61` at `sample[0].set[0].par` and no value at the new path. These observations become null and are discarded, silently suppressing the sensor line or excluding those instruments. Flatten the supported sample/set entries before aggregating.

### P2

3. **“Last 30 days” selects 30 activity rows.** [activity-chart.tsx](https://github.com/Jan-IngenHousz-Institute/open-jii/blob/5c36fde5e3af91578f9d747fb8244bef40dc16d0/apps/web/components/public-metrics/activity-chart.tsx#L41) slices a sparse series without calendar bounds. Thirty weekly rows span 203 days. The cumulative view also anchors to the latest activity rather than today. Filter by calendar dates and fill silent days.

4. **UTC midnight can corrupt active counts and the busiest resource.** [Resource aggregation](https://github.com/Jan-IngenHousz-Institute/open-jii/blob/5c36fde5e3af91578f9d747fb8244bef40dc16d0/apps/backend/src/metrics/application/resource-metrics.service.ts#L124) treats every date outside the previous window as current; scoped aggregation does too. A ten-minute cache can retain a date now outside both windows. A reproduction returned total `2` but busiest-resource count `99`, because densification drops the old date from totals while the resource map retains it. Explicitly test membership in the current window.

5. **Experiment summaries include archived experiments hidden by the default list.** [getVisibleExperimentIds](https://github.com/Jan-IngenHousz-Institute/open-jii/blob/5c36fde5e3af91578f9d747fb8244bef40dc16d0/apps/backend/src/metrics/core/repositories/metrics.repository.ts#L102) checks access but omits the default archive exclusion. Counts and the busiest card can therefore describe experiments absent below. Share the archive policy or explicitly present the summary as including archived activity. This is not an authorization leak.

6. **The derived caption labels a 90-day ranking “this month”.** [The translated caption](https://github.com/Jan-IngenHousz-Institute/open-jii/blob/5c36fde5e3af91578f9d747fb8244bef40dc16d0/packages/i18n/locales/en-US/publicMetrics.json#L27) conflicts with `PARAMETER_WINDOW_DAYS = 90`; German and Dutch do likewise. State the actual window, ideally using contract metadata.

7. **Milestone dismissal resets on remount.** [milestone-banner.tsx](https://github.com/Jan-IngenHousz-Institute/open-jii/blob/5c36fde5e3af91578f9d747fb8244bef40dc16d0/apps/web/components/dashboard/milestone-banner.tsx#L21) keeps dismissal only in component state, so the same banner returns after navigation or reload and can remain eligible for months. Persist dismissal by milestone identity.

### Smaller confirmed defects

- [Mean arrival gap](https://github.com/Jan-IngenHousz-Institute/open-jii/blob/5c36fde5e3af91578f9d747fb8244bef40dc16d0/apps/data/src/pipelines/metrics/pool_facts.py#L81) divides by `arrivals` rather than `arrivals - 1`: two samples 60 seconds apart yield 30 seconds.
- [Parameter descriptions](https://github.com/Jan-IngenHousz-Institute/open-jii/blob/5c36fde5e3af91578f9d747fb8244bef40dc16d0/apps/web/components/public-metrics/parameter-line.tsx#L22) interpolate English pipeline labels into German/Dutch sentences. Localize by stable parameter name.
- [Macro summary copy](https://github.com/Jan-IngenHousz-Institute/open-jii/blob/5c36fde5e3af91578f9d747fb8244bef40dc16d0/apps/web/components/metrics/resource-metrics-summary.tsx#L55) and shared trends/row strips call macro execution counts “measurements”. Use kind-specific units.
- [Dashboard documentation](https://github.com/Jan-IngenHousz-Institute/open-jii/blob/5c36fde5e3af91578f9d747fb8244bef40dc16d0/apps/docs/content/guide/get-started/platform-tour.mdx#L53) promises a last-day platform figure that the dashboard activity panel does not render.

### Validation limits

Reviewers reported passing application tests, type/build checks and lint; focused reproductions confirmed the blocking, date-window and arithmetic failures above. Spark validation remains incomplete, and no live browser or warehouse verification was performed. Node/JDK version mismatches affected local validation. No new authorization leak was confirmed. Unverified Ambit vocabulary concerns and cosmetic preferences were excluded from these findings.

petar-INFO:
@dominikvrbic addressed, thx

PR #2011, opened 2026-08-31
feat(mobile): EAS Observe + Insights instrumentation, expo agent skills

dominikvrbic:

- [x] Feature
- [x] Chore

## Description

Wires up two EAS telemetry surfaces ahead of the next store release:

1. **EAS Observe** — `expo-observe` installed; `RootLayout` wrapped with `ObserveRoot.wrap()` (SDK 56+ API) and `markInteractive()` called where the app genuinely becomes usable: after the force-update gate and navigation are ready, alongside the splash hide. Gives production startup metrics (TTR, TTI, cold/warm launch) and, on SDK 57, JS error reporting.
2. **EAS Insights app usage** — `expo-insights` installed; reports cold starts automatically, no wiring needed. The App usage dashboard feeds the OJD-1473 adoption baseline (installs/usage over time, by store version).
3. **`expo-skill-lookup`** — a single router skill (`.agents/skills/`, symlinked into `.claude/skills/`) that tells agents how to fetch the right upstream Expo/EAS skill on demand (`pnpm dlx skills use expo/skills@<name>` or raw GitHub). The 26 upstream skills are deliberately not vendored.

**Release note:** both libraries are native modules, so the next release requires a **full build** — an OTA cannot carry them. `app.json` already has the EAS `projectId` linkage both need.

## Related Issues

- Contributes to OJD-1473

## Testing Instructions

- [x] Full mobile suite: 107 files / 1124 tests pass; `tsc` and eslint clean
- [x] Local release build compiles with the new native modules (arm64)
- [ ] Device smoke test pending — phone disconnected mid-session; verify app boots and an Observe session appears in the EAS dashboard

## Changes Made

- `apps/mobile/src/app/_layout.tsx`: `ObserveRoot.wrap(RootLayout)`, `useObserve().markInteractive()` in `GatedApp`
- `apps/mobile/package.json`: `expo-observe ~57.0.17`, `expo-insights ~57.0.14`
- `.agents/skills/`, `.claude/skills/`, `skills-lock.json`: expo skills

## Checklist

- [x] I have updated the documentation accordingly

## Additional Notes

markInteractive is placed at the single startup gate both entry screens (login and tabs) pass through; per the docs it is safe to call multiple times and only the first records. Debug builds do not dispatch metrics by default, so verification is via a release build and the Observe dashboard.


petar-INFO:
## Via Opus

**Request changes.** The mobile code is small and correct; the gaps are verification and a couple of small hardening items.

Verified sound first: `runtimeVersion.policy: "fingerprint"` is the policy that hashes native dependencies, so the commit's "no OTA" claim is enforced by config rather than discipline. Neither `expo-observe` nor `expo-insights` ships an `app.plugin.js`, so no `plugins` entry is needed. Mobile `tsc --noEmit` is clean. The skill router's URLs resolve, and upstream confirms `ObserveRoot` + `useObserve()` is the right API for SDK 56+ (this app is on 57).

**1. Nothing shows the metrics actually arrive.** For an observability feature that is the deliverable, and it is the one thing CI cannot check. Both packages are native, so it needs a dev build on a device rather than Expo Go or an OTA update.

Related, and it changes what is being collected: the app never calls `Observe.configure(...)`, so `isExpoRouterInitialized()` is false and `useObserve()` falls through to the raw `AppMetrics.markInteractive`. You get the app-level cold-start metric, not per-route TTR/TTI. If navigation timing was part of the goal it is not wired; if app-level was the goal, the subject reads broader than the change.

**2. Make the interactive marker fire-once.** Two separate questions both disappear with one guard. First, `initialUiReady` is `forceUpdateStatus === "gated" || (forceUpdateStatus === "allowed" && navigationReady)`, and whether that can flip more than once depends on whether `ForceUpdateGate` re-checks; if it can, `SplashScreen.hideAsync()` and `markInteractive()` both re-run today. Second, the dep is stable only because no integration is configured: enable the router integration (item 1) and `useObserve()` returns `useObserveForRouter`'s `useCallback`, keyed on `routeParams`, `navigation`, `pathname` and `storage`, so the dep churns on every navigation. The package guards duplicate TTI emission per screen id, but `AppMetrics.markInteractive` is invoked before that guard.

The package ships the primitive for this: `<ObserveInteractiveMarker />`, documented as calling `markInteractive` once on mount. Use it, or gate the effect body on a ran-once ref.

**3. The `dnf` hunk has no rationale, and it departs from the convention in its own directory.** `functions/r/Dockerfile` names its CVEs and the fixed version inline, and both `.trivyignore.yaml` files carry package@version, fixed-in version, a statement and an `expired_at` for every entry. Those files also state the governing philosophy: AWS-managed components wait for an AWS base republish. This upgrades an AWS-managed package at build time instead, with no comment saying which finding it clears or why openssl warrants that when gawk, glib2 and rpm are suppressed and left to AWS. One comment matching the neighbours settles it. Separately, `--releasever=latest` makes the layer non-reproducible where the surrounding practice pins versions.

**4. Telemetry disclosure, which needs a decision rather than a code change.** Web PostHog is explicitly consent-gated for GDPR (`getConsentStatus()`, `cookieless_mode: "on_reject"`, memory-only until opt-in). Mobile has no gate and already sends `captureAppLifecycleEvents`, so Insights is not a new category there. What is new is the processor: EAS is a different controller from PostHog, and `expo-insights` exports `{}`, so there is no runtime surface to gate or disable it, now or later. If the Contentful-backed policies page enumerates processors, it likely needs EAS added.

**Worth splitting, your call:** the Lambda OpenSSL fix in `f8da067f8` is unrelated to mobile observability, and bundling it means the security fix waits on this review and cannot be reverted independently. I checked before pressing it: the nearest precedent went the other way (the R libgnutls30 CVE fix landed inside a measurement-upload feature PR), and `deploy-macro-sandbox.yml` is manual-dispatch only, so nothing auto-deploys. So this is reviewer preference rather than a convention you are breaking.

**Mechanical:**
- The branch does not merge cleanly: `pnpm-lock.yaml` conflicts with the two dependency PRs on main. Rebase needed regardless.
- PR metadata per the CONTRIBUTING contract: conventional-commit title plus a **Linear issues** section. OJD-1473 reads like `Contributes to`, not `Closes`. The skill and Dockerfile commits need covering too, or the `no-linear` label with a note.

PR #2032, opened 2026-09-08
feat(api,db): iot calibration contract, capture-procedure schema, and tables

petar-INFO:

- [x] Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Test Update
- [ ] Chore
- [ ] Other (please describe)

## Description

The contract and storage for a device-calibration domain, defined before either side implements it. A calibration definition is a versioned recipe naming a sensor family and carrying three things: a capture procedure, a computation script, and an output schema of coefficient specs whose bounds are checked when a result is produced and again at approval. A definition also declares a minimum firmware version and refuses runs on devices below it, so a procedure issuing commands the firmware does not know cannot return numbers that merely look like data.

The capture procedure is a declarative, fully validated description of a bench session rather than free text. It declares a rig of instruments and a sequence of operator, settle, read, and sweep steps; a read produces named columns and a sweep produces a series with a reserved stimulus column and one column per read. The schema enforces the integrity the bench depends on: unique instrument roles and series names, no duplicate columns, every referenced instrument declared, and optional steps a bench may legitimately skip. Series, role, and column names share one lowercase-identifier shape because the same string becomes a payload key, a Python dict key, and a DataFrame column.

Three tables back this (definitions, runs, applied calibrations) via migration 0050, and the branch also repairs the 0049 meta snapshot so the next `db:generate` does not emit unrelated churn.

## Linear issues

Contributes to OJD-1602


dominikvrbic (changes requested):
## Stack review: #2032 (contract + schema)

Part of a coordinated review of the 7-PR calibration stack (#2032 → #2104) run as ten independent reviewer passes (Fable, Opus, GPT-5.6 Sol, GPT-6 Astra, Kimi K3), each validating its findings against the code before reporting. Findings are posted on the PR whose diff should carry the fix. Stack head reviewed: `54d3d8ed1`.

**Verified here:** `@repo/api` and `@repo/database` typecheck, 1,897 API tests pass, `db:generate` emits no churn, the 0049 snapshot repair is limited and correct, and the full drizzle chain 0000→0051 applies cleanly on PostgreSQL 16.15 (prod is Aurora PG 16.11). A reviewer initially flagged the table-qualified reference in 0050's generated column as a Postgres error; that was refuted by running the migration and is not a finding.

1. **Definition versioning has no identity or uniqueness.** `packages/api/src/domains/iot/calibration/iot-calibration.contract.ts:26-42`, `packages/database/src/schema.ts:940-972`. The contract says posting an existing `(family, name)` creates a new version, but the create body carries no lineage id, `version` defaults to 1, and no unique index ties versions to a series. Two creates for the same pair produce two unrelated rows both reporting version 1. Pick the series identity, allocate the next version transactionally, and enforce it in the DB.

2. **Output schemas accept no-op and impossible calibrations.** `iot-calibration.schema.ts:15-59, 385-453`. A block's coefficient map may be empty and every spec accepts `min > max`. `{ blocks: { par: {} } }` plus a computed block with `coefficients: {}` yields no violations, `hasComputedBlock()` is true, and an applied `par` block that writes nothing is created. Require at least one coefficient per block and refine `min <= max`.

3. **Step bound and series bound never reconciled.** `procedure.schema:181` allows 64 steps; `iot-calibration.schema.ts:203-207` allows 20 payload series, and every step may add a `_retaken` companion. An 11-step procedure with one retake per step is refused at submit, after the bench session. Check the two bounds against each other at definition creation.

4. **Payload row keys bypass the declared-column guarantees.** Series row keys are not constrained to the procedure's declared, lowercase column identifiers, so the "same string becomes a payload key, a dict key, and a DataFrame column" invariant only holds by convention.

5. **Firmware strings accepted by zod exceed the DB columns.** `min_firmware_version` is `varchar(32)` and the device firmware column is bounded too, while the zod schema accepts longer strings. Insert fails after validation passed.

6. **A write report may omit every block** and still satisfy the contract, so an "applied" row can exist with nothing confirmed written.

### P2

- `zSeriesCell` caps strings at 4096 characters (`iot-calibration.schema.ts:194-200`). The interpreter in #2103 JSON-stringifies structured device replies into cells, and MultispeQ/MiniPAR envelopes are tens of KB. The fix is discussed on #2103, but the cap itself needs a decision: either raise it for structured cells or forbid structured cells in the contract.
- No `apps/docs/content` change anywhere in the stack. The capture-procedure grammar (roles, handshakes, setpoint placeholders, reserved `stimulus` column, `_retaken` convention) has no written reference to author against. AGENTS.md makes docs a same-PR requirement.

petar-INFO:
Thanks, this was a genuinely useful pass. Every finding was checked against the code before acting on it; below is what changed and what I am pushing back on. The stack has been rebased, so all seven branches have new heads.

### Fixed here

**2. No-op and impossible output schemas.** `zCoefficientSpec` now refines `min <= max` on every variant, and `zCalibrationOutputSchema` refines that each block declares at least one coefficient. `{ blocks: { par: {} } }` is now refused at definition creation rather than producing an applied block that writes nothing. The authoring UI follows: adding a block by hand starts it with one coefficient, so the author never gets a block that cannot be saved (on the library branch).

**3. Step bound and series bound reconciled.** `MAX_PROCEDURE_STEPS = 64` and `MAX_VERIFY_STEPS = 16` are exported from the procedure schema and used by `zCaptureProcedure`, and `MAX_PAYLOAD_SERIES` is now derived as `MAX_PROCEDURE_STEPS * 2` so that a retake on every step still fits. The two bounds can no longer drift apart.

**5. Firmware strings within the column.** `zFirmwareVersion` is capped at 32 to match `min_firmware_version varchar(32)`.

While fixing this I hit the other half of the same problem, raised as finding 8 on #2104: field firmware carries release suffixes, and the anchored schema made the wizard drop such a version rather than send it, after which the floor check refused the run for "reporting no version". There is now a second schema, `zReportedFirmwareVersion`, used by the two create bodies. It accepts what a device answers, suffix and all, still within the column, while `zFirmwareVersion` stays anchored for the floor an author declares. The run record keeps the device's own words rather than a truncated core.

**6. Empty write report.** `zReportDeviceCalibrationWriteBody.writeResults` is refined non-empty, so a report cannot stamp a calibration as written while confirming nothing.

**P2, the cell cap.** Raised to 65,536 characters (`MAX_SERIES_CELL_TEXT`). The rest of that finding is answered on #2103: the interpreter now refuses an oversized cell at the read, so the operator hears about it at the step that produced it rather than as a 400 after the whole session.

### Fixed, but on #2033

**4. Payload row keys.** The guarantee now holds by construction rather than by convention: `payloadColumnIssue` constrains each row's keys to the columns its step declares (`read.as`, plus the reserved `stimulus` for a sweep), and a `_retaken` companion is held to its series' columns. It landed on the domain branch because that is where the procedure-versus-payload checks already live.

### Not taking

**1. Definition versioning.** This one rests on a premise I should have written down more plainly, and the contract comment that suggested otherwise has been corrected. A definition's name is a label, as it is for a protocol or a workbook, not a key: two definitions may share a name and are unrelated rows. Nothing posts "a new version of this name". `version` exists so that a run records which revision of a definition produced it, and since a definition is frozen the moment it has a run, it stays at 1 for the whole of its life. There is no lineage to allocate and nothing for a unique index to protect. If we later want an explicit "supersede this definition" action, that wants a real parent id on the row and its own endpoint, not an inference from the name.

### Deferred by an explicit decision

**P2, docs.** The calibration documentation is being written once, at the end of the programme, rather than accreting across eight intermediate PRs that redefine the same grammar as they go. That is a deliberate call for this stack and not an oversight.

### Where it landed

- [`cbe9f6b12`](https://github.com/Jan-IngenHousz-Institute/open-jii/commit/cbe9f6b12) refuse empty blocks, inverted bounds, overlong firmware and empty write reports, and bound the payload by the procedure
- [`8ae758b05`](https://github.com/Jan-IngenHousz-Institute/open-jii/commit/8ae758b05) record the firmware a device reports, release suffix and all, instead of dropping it
- [`f7e29cfb1`](https://github.com/Jan-IngenHousz-Institute/open-jii/commit/f7e29cfb1) regenerate the api spec for the calibration contract's bounds

The whole stack is rebased onto `main` (`8be3155b3`) and force-pushed, so every branch carries the seven commits main gained this week, and the generated API spec is regenerated on each branch that changes the contract so the drift gate passes per PR. Workspace gate on the stack head: web 6,757 tests across 947 files, backend 3,498 across 299, `@repo/api` 1,930, `@repo/iot` 604, plus lint, typecheck and the production web build. CI on this PR: 23 checks green.

dominikvrbic (approved):

PR #2033, opened 2026-09-08
feat(backend): iot calibration domain with runs, approval flow, and calibration_definition resource type

petar-INFO:

- [x] Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Test Update
- [ ] Chore
- [ ] Other (please describe)

## Description

Implements the calibration domain the contract defines, and makes a calibration definition a first-class shareable resource. A run is scored, not trusted: creating one from the bench wizard resolves the definition and device, checks the device firmware meets the definition minimum, and verifies the captured payload carries every required series and nothing the procedure cannot produce, then records the run and invokes its script. A second path records blocks an external bench computed itself, without running a script. Both meet the same gate.

Approval is that gate, and it is human: nothing a script produced reaches a device until someone with device manage rights approves it. Approval re-validates the blocks against the definition's output-schema bounds, then in one transaction supersedes the device's in-force calibration and writes the applied row from only the computed blocks, leaving rejected and skipped ones on the run as the record of what the session could and could not do. Rejection is terminal and keeps the diagnostics.

`calibration_definition` runs through the access-control statements and roles, the sharing resource-type set, resource staffing, visibility, and the organization resources listing, so a definition shares, publishes, and lists like any other resource. Migration 0051 adds the enum value on top of the tables the contract branch created.

## Linear issues

Contributes to OJD-1602


dominikvrbic (changes requested):
## Stack review: #2033 (backend domain, approval, resource type)

**Verified here:** the approval transaction holds under a real two-approval race against the partial unique index; re-validation at the gate; applied row built from computed blocks only; newest-block-per-name assembly; no N+1 in the domain; approval requires device manage so there is no cross-org approve without an explicit grant; visibility and publish match protocols/macros/workbooks; every exhaustive resource-type list was updated (warehouse metrics excluded deliberately); migration 0051's `ALTER TYPE ADD VALUE` is transaction-safe on PG 16; backend typechecks clean.

1. **External-run path bypasses the firmware floor.** `create-external-calibration-run.ts:43-100` has no min-firmware check and approval never re-checks it, while both run endpoints are gated identically on `device:contribute`. The exemption is described as a historical-import path but is not scoped to importing: the same operator can post to the sibling URL to skip the floor. Add an explicit `historical: true` waiver, or surface a below-floor marker at approval.

2. **Definition delete skips the grant protocol.** `iot-calibration-definition.repository.ts` (delete) runs without a transaction, `lockStaffedResource`, or `deleteResourceGrants`, unlike every other staffed resource type. Leaves orphaned `resource_grants` rows and races concurrent share writes.

3. **Every definition link 404s.** `calibration_definition` is wired into global search (`command-palette.tsx:197`), org resource rows (`organization-resource-meta.ts`) and sharing routes (`resource-routes.ts`) under segment `calibrations`, but no `/platform/calibrations/[id]` route exists anywhere in the stack, including #2104. Either add the page in #2104 or don't register the segment until it exists.

4. **Sandbox traceback dropped.** `create-calibration-run.ts:184`. `traceback` is declared in `zCalibrationSandboxResponse`, sent by the sandbox, and documented at `packages/database/src/schema.ts:935`, but `invokeAndSave` persists only `error` and `reasons`. Script authors never see the failing line, and a definition cannot be edited once it has runs.

5. **List endpoint loads every run's full payload.** `iot-calibration-run.repository.ts:96`. `listByDevice` selects and zod-parses each run's readings (up to 20 series × 5000 rows) and then `zCalibrationRunList` strips them on the wire, on every device page load. Project the payload out of the list query.

6. **No device identity check before write-back** (root cause is in #2104, but the backend side belongs here). Nothing records which physical device identity received the write. Have the write-report endpoint carry the identified device serial and reject a mismatch with the route device.

### P2

- Losing the approve/reject race returns 500, not 409: raw `throw new Error("This run was already reviewed")` falls through `defaultRepositoryErrorMapper`.
- A run stuck in `running` or `error` can never be closed: reject accepts only `computed`.
- `update-calibration-definition`'s run-count edit freeze is check-then-act outside a transaction; a concurrent run creation slips in.
- `create-calibration-run.ts:131-140` passes `*_retaken` series to the sandbox, contradicting the contract's "the fit never sees it".

petar-INFO:
Thanks. All six P1s and all four P2s are addressed; the branch has a new head after a stack-wide rebase.

### Fixed

**2. Definition delete now follows the grant protocol.** The delete runs in a transaction, takes `lockStaffedResource(tx, "calibration_definition", id, "update")`, calls `deleteResourceGrants`, and only then removes the row, matching every other staffed resource type. No more orphaned grants and no race with a concurrent share write.

**4. Sandbox traceback kept.** `invokeAndSave` folds the traceback into the run's `errorMessage`, so the failing line survives on a definition that is already frozen by its first run. Pinned by a test.

**5. List endpoint no longer carries payloads.** `listByDevice` projects the payload out of the query with `getTableColumns` minus `payload`, and returns `CalibrationRunSummaryDto[]`. The readings still travel with the single-run detail route, which is the one place they are wanted.

**6. Device identity.** Both `zCreateCalibrationRunBody` and `zReportDeviceCalibrationWriteBody` now carry an optional `reportedSerial`, and the contract exports `serialsMatch`, which compares on alphanumerics only so that a MAC printed with or without separators still matches the registered serial. Creating a run and reporting a write both refuse a mismatch, naming both serials. The field is optional because MiniPAR firmware announces no unit at all; the browser side (#2104) says so plainly on the connect step rather than letting silence read as a match.

**P2, the approve/reject race returns 409.** The raw `throw` is now `AppError.conflict("This run was already reviewed")` inside the transaction.

**P2, stuck runs can be closed.** Reject accepts `computed`, `compute_failed` and `error` (`REJECTABLE_STATUSES`); `running` still refuses, since a run in flight has not finished doing whatever it is doing.

**P2, the edit freeze is no longer check-then-act.** The run-count guard moved into the UPDATE statement as a `notExists` on runs for that definition, so a run created concurrently cannot slip between the check and the write.

**P2, `_retaken` is filtered.** Only the series the fit is meant to see reach the sandbox event; the retaken companions stay on the run record, which is what the contract says.

Also landed here, from #2103's finding 6: `skipped_series` as a jsonb column with migration 0052, `zSkippedSeriesList` on the run and the create body, and the row-key constraint from #2032's finding 4 (`payloadColumnIssue`, plus `verificationSeriesIssue` for the verify phase).

### Context rather than a change

**1. External-run path and the firmware floor.** The exemption is deliberate and is now documented at the use case rather than left implicit. This endpoint records a fit a bench tool computed itself, including sessions from before the device's firmware was tracked at all, so a floor that refuses a run for reporting no version would reject exactly the imports it exists to accept. The floor protects the opposite case: a run the platform drives, where an unknown command would return numbers that look like data. I would rather not add a `historical: true` waiver that every caller sets. If we want a stronger signal, the honest one is a below-floor marker surfaced at approval, and I am happy to take that as a follow-up once someone is actually importing.

**3. Definition links.** The `/platform/calibrations/[id]` route exists, on `feat/calibration-library`, which is the eighth branch of this stack and does not have a PR open yet. The segment registration and the page land together; nothing in the seven open PRs ships a link without its target.

### Where it landed

- [`60996fce2`](https://github.com/Jan-IngenHousz-Institute/open-jii/commit/60996fce2) hold payload rows to their declared columns, record the steps a bench skipped, and compare the unit's own serial
- [`cdcec0d21`](https://github.com/Jan-IngenHousz-Institute/open-jii/commit/cdcec0d21) refuse a session on the wrong unit, keep the traceback, close failed runs, delete a definition's grants, and race-proof the edit freeze
- [`f943c5fbc`](https://github.com/Jan-IngenHousz-Institute/open-jii/commit/f943c5fbc) pin that a firmware version with a release suffix passes a definition's floor
- [`11be262dc`](https://github.com/Jan-IngenHousz-Institute/open-jii/commit/11be262dc) regenerate the api spec for the skipped-series and reported-serial fields

dominikvrbic (approved):

PR #2035, opened 2026-09-08
feat(sandbox): isolated runtime for calibration scripts with shared QC gates

petar-INFO:

- [x] Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Test Update
- [ ] Chore
- [ ] Other (please describe)

## Description

The runtime that turns a definition's script into coefficients. One generic Python Lambda takes the script, the captured series, the run parameters, and the output schema per synchronous invoke, turns each series into a DataFrame, and injects exactly three names into the script: `inputs`, `params`, and a `submit` callback. Whatever the script submits is validated against the output schema and a shared set of QC gates before anything returns, and the result carries a status: computed for validated blocks, compute_failed when the script is at fault, error for a malformed event.

Isolation is deliberate and layered: an isolated Lambda with zero egress at the boundary, plus a fresh process per run, rather than in-process exec hardening. That rests on an explicit trust model, definitions are versioned reviewable artifacts and human approval is the security gate, so the in-process validation catches honest mistakes rather than a hostile script. The QC gates live in one module so two scripts fitting the same kind of curve cannot disagree about what passes, and a local runtime-emulator endpoint lets a script be authored against an exported bench payload with no AWS in the loop.

## Linear issues

- Contributes to OJD-1602


dominikvrbic (changes requested):
## Stack review: #2035 (Python sandbox runtime)

Part of a coordinated review of the 7-PR calibration stack (#2032 → #2104) run as ten independent reviewer passes (Fable, Opus, GPT-5.6 Sol, GPT-6 Astra, Kimi K3), each validating its findings against the code before reporting. Stack head reviewed: `54d3d8ed1`. A second, independent review of this stack earlier the same day reproduced two additional sandbox-escape findings; they are included below with attribution.

**Verified here:** all 97 Python tests pass on the exact Dockerfile pins (numpy 2.4.2, pandas 3.0.1, scipy 1.17.0); each run really is a new interpreter with a rebuilt child env, read-only params, a 25 s alarm inside a 30 s outer timeout; DataFrames preserve row and column order; coefficient checks (names, finiteness, array length, integer, inclusive bounds) behave; QC helpers fail closed on constant data, two-point fits, NaN and inf; the event shape matches what #2033 sends.

### P0, blocks deploy

1. **Backend fails to boot when the emulator endpoint is unset.** This PR adds `apps/backend/src/common/modules/aws/services/config/config.service.ts:67-69`, which loads `aws.lambda.calibrationSandboxEndpoint` with `getOrThrow`. `aws.config.ts:27` maps it from `AWS_LAMBDA_CALIBRATION_SANDBOX_ENDPOINT`, which the README says to leave unset to use the deployed Lambda. Nest's `getOrThrow` throws on `undefined` (verified against the installed ConfigService), so every AWS backend task crashes at startup. `.env.test` sets the var to an empty string, which passes, so no test catches it. #2101's task definitions do not inject it. Fix: `get(...) ?? ""` and a test with the variable genuinely absent.

### P0, trust boundary (reproduced by the parallel review)

2. **Allowlisted pandas reaches credentials, network and code execution.** `pd.read_csv("/proc/1/environ", sep="\0")` returns the Lambda role's `AWS_SECRET_ACCESS_KEY` and session token (same uid as the runtime process, so stripping the child env does nothing). `pd.read_csv("http://...")` reaches the network. `pd.read_pickle` over a URL executes code. `handler.py:6-9`, `wrapper.py:44`, the README and `test_isolation.py:167` all claim otherwise; the isolation test only checks `open` and `import tempfile`. The only real containment is #2101's IAM deny plus VPC egress rules, and that security group is open to the whole VPC CIDR on 443.

3. **QC module can be poisoned for later runs in a warm container.** The container runs as uid 0 with the source tree writable in the local image; a pandas-only script overwrote `helpers/qc.py`, and every later run imported the forged gate. Reproduced locally: unfittable data went `rejected` with four reasons, then after one poisoning run the same data returned `computed`, `passed: true`, no reasons, fabricated coefficients. Real Lambda mounts the image read-only except `/tmp`, so confirm on a deployed function before treating this as AWS-reachable; finding 4 holds regardless.

Fix for 2 and 3: non-root uid, read-only filesystem, no network at the OS level, and a test that actually writes to disk. If the intended model is "trusted reviewed authors only", say so in the README and enforce it before execution; approval after computation does not protect the runtime.

4. **Warm filesystem shared across runs.** `handler.py:226` changes `HOME`/`TMPDIR` but restricts nothing. A script wrote a CSV via `DataFrame.to_csv` and the next invocation read it back through `pd.read_csv`; both returned `computed`.

5. **Output limits applied after unbounded buffering.** `handler.py:218` uses `subprocess.run(capture_output=True)`; the 10 MiB stdout check at line 244 runs after collection, stderr has no cap, and `wrapper.py:304` captures all script printing in an unbounded `StringIO`. A chatty script OOMs the 1024 MiB Lambda instead of yielding `compute_failed`. Stream with byte counters and kill the worker at the limit.

### P2

- `handler.py:97` accepts block metadata the API rejects: `reason` type/length (max 2000) and 16 KiB fit/quality records per `zCalibrationBlock` (`iot-calibration.schema.ts:77`). Three benign probes returned `computed` from Python and failed the API schema; the backend stores `error` with "unrecognized payload" and the explanation is lost.
- `wrapper.py:331` serializes after the script-error try block, so a cyclic submission becomes `status: error`; `handler.py:104` does set membership on an unchecked `status`, so `status: []` raises TypeError. Both are script faults misreported as infrastructure errors.
- Docs vs behavior: QC gates are advisory in code and tests (`test_failed_qc_gates_are_advisory_and_travel_with_the_block`), while the README and PR description say a failed gate rejects a block. Scripts also receive `np`, `pd`, `scipy`, `qc` and allowlisted builtins, not "exactly three names". Pick one and fix the other.

petar-INFO:
The two escape findings were right, and they were right about the README too: it described a boundary the code did not have. Everything below is fixed and verified in the container, and the same hardening was applied to the macro sandbox, which had the identical holes.

### P0

**1. Backend boot.** `calibrationSandboxEndpoint` is read with `get<string>(...) ?? ""` instead of `getOrThrow`, and there is a test that boots with the variable genuinely absent rather than set to an empty string, which is what let the old one pass.

**2 and 3. pandas reaching credentials, network and code execution; QC poisoning.** A Python audit hook (PEP 578, `sys.addaudithook`) is installed after the script is compiled and before it is executed, so it covers the script and everything the script calls, including allowlisted libraries reaching through C. It refuses `open` for write in any mode, and for read outside a small set of roots (the interpreter's own prefix and the helpers directory), and it refuses the events behind sockets, urllib, http.client, the other network clients, subprocess, `os.system`/`exec`/`spawn`/`fork`, the os mutation calls, shutil, tempfile, pty, ctypes, webbrowser and sqlite3. `pd.read_csv("/proc/1/environ")`, `pd.read_csv("http://...")`, `pd.read_pickle` over a URL and writing over `helpers/qc.py` all now raise a `PermissionError` that the run reports as a script failure.

The process also runs as an unprivileged uid. The Lambda Python image has no `useradd`, so the account is appended to `/etc/passwd` and `/etc/group` directly and the image sets `USER 1001`.

The README no longer claims the interpreter is the boundary. It now says what is true: the boundary is the function's execution role and its security group, the audit hook and the unprivileged user are layers inside it, and a script is reviewed before it is published. `LibraryEscapeTest` covers the reaches above and fails against the old code.

### P1

**4. Warm filesystem.** Same hook. A script cannot write a file for a later invocation to read, in `/tmp` or anywhere else.

**5. Output limits.** `subprocess.run(capture_output=True)` is replaced by a `Popen` with a draining thread per pipe, each with a byte counter; passing the cap kills the child and the run returns `too_large` rather than taking the 1024 MiB Lambda down with it. stderr has its own cap. The wrapper's `StringIO` is now a `CappedBuffer` that stops accumulating past 64 KiB.

### P2

- Block metadata is validated where the API validates it: `reason` must be a string of at most 2000 characters and each record must serialise within 16 KiB, so a benign probe now comes back as a script failure that explains itself instead of a stored "unrecognized payload".
- `_jsonable` moved inside the script-error try block, so a cyclic submission is a script failure rather than an infrastructure error, and `status` is checked with `isinstance(status, str)` before the set membership that used to raise on `status: []`.
- Docs versus behaviour: the README now matches the code on both counts. A failed gate rejects its block, which is what the bench tool this replaces does, and the names injected into a script are listed rather than described as "exactly three".

### Macro sandbox

Raised separately in the thread and worth repeating here: `apps/macro-sandbox` had the same escapes, the same root user and the same unbounded buffering, so it got the same hook (installed before the batch loop), the same streaming handler, the same unprivileged uid and the same README correction. Verified against the runtime emulator: mean over a DataFrame still works, pandas read and write and a numpy save are refused, numpy maths is untouched.

All 109 Python tests pass in the container.

### Where it landed

- [`28cc352f1`](https://github.com/Jan-IngenHousz-Institute/open-jii/commit/28cc352f1) boot without a calibration sandbox endpoint instead of crashing every deployed task
- [`fd8603602`](https://github.com/Jan-IngenHousz-Institute/open-jii/commit/fd8603602) refuse library escapes at the interpreter, bound output as it arrives, run unprivileged, and say what the boundary is, in both sandboxes

dominikvrbic (approved):

PR #2078, opened 2026-09-14
fix(web): replace the broken latest-measurements card with an experiment data inventory

petar-INFO:

- [ ] Feature
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Test Update
- [ ] Chore
- [ ] Other (please describe)

## Description

The experiment overview's latest-measurements card was broken. It hardcoded the device table and rendered the four newest raw rows, so it showed nothing for experiments whose data lands anywhere else and, where it did render, a handful of raw rows told you little about what the experiment actually holds.

It's replaced with a data inventory: every table the experiment has, each with its type and total row count, and a link through to the Data tab. That pairs with the activity pulse next to it, the pulse says whether data is still arriving, the inventory says what has accumulated, which is the one thing the overview otherwise can't tell you without opening the Data tab. Loading, empty, and error states are handled, and the archived overview gets the same card pointed at its own data route.

## Linear issues

- Contributes to OJD-1768


apps/web/components/experiment-overview/experiment-data-inventory-row.tsx
```diff
@@ -0,0 +1,35 @@
+"use client";
+
+import { useLocale } from "~/hooks/useLocale";
+
+import type { ExperimentTableMetadata } from "@repo/api/domains/experiment/data/experiment-data.schema";
+import { useTranslation } from "@repo/i18n";
+import { Badge } from "@repo/ui/components/badge";
```
selianaaaa, line 7: finally a project where people care about imports sorting

apps/web/components/experiment-overview/experiment-data-inventory.tsx
```diff
@@ -0,0 +1,98 @@
...
+  );
+}
+
+interface DataInventoryBodyProps {
+  isLoading: boolean;
+  hasError: boolean;
+  tables: ExperimentTableMetadata[];
+}
+
+function DataInventoryBody({ isLoading, hasError, tables }: DataInventoryBodyProps) {
```
selianaaaa: aren't we following a one component per file pattern in the web?

selianaaaa (approved):
Looks good to me

apps/web/components/experiment-overview/experiment-data-inventory-row.tsx
petar-INFO: linter enforces it 😄

apps/web/components/experiment-overview/experiment-data-inventory.tsx
petar-INFO: Yes up to a pragmatic degree .. but I will merge this one with its parent now that I look at it again

PR #2081, opened 2026-09-14
feat(mobile): keep protocol/macro snapshot code out of the persisted measurement-flow slice

vladstoe:
## Summary

The measurement-flow store persists `flowNodes`, and until now every node carried its protocol JSON and macro source copied in from the workbook version's entity snapshots. Zustand `persist` re-serialises the whole slice on every `set`, so each scan result, step change or branch visit rewrote every protocol and macro body in the workbook to AsyncStorage, even though that code never changes for a pinned version.

This PR strips the code from the persisted copy and re-attaches it on a cold-start resume, behind a loading gate, so a resumed measurement node never renders with code it cannot run. The code is written once per flow to its own small store, so offline resume does not depend on the query cache at all.

## What changed

- `partialize` now maps `flowNodes` through `stripSnapshotCode`, which removes `content.protocol.code` and `content.macro.code` and nothing else. In-memory nodes keep the code, so every consumer reads it exactly as before.
- New `hasUnresolvedSnapshotCode` predicate: a snapshot object with `code === undefined` means "stripped, not yet re-hydrated". Resolved-empty code (`[]` / `""`, a version with no snapshot) is not unresolved and never gates.
- New store action `rehydrateFlowNodes` replaces `flowNodes` without resetting progress. `setFlowNodes` and `setFlowGraph` both reset, which would wipe the resumed position.
- New `useFlowSnapshotsStore` (`measurement-flow-snapshots-storage`, persist v1) holding `{ workbookVersionId, entitySnapshots }` for the active flow. `useLoadExperimentFlow` writes it once when the flow loads; `teardownFlow` and the rehydration guard clear it. Its wire format is pinned in `flow-store-persistence.test.ts` like the other two flow stores. This is what keeps offline resume independent of the query cache: a `buster` bump on app update or an eviction drops the query cache, but this store is versioned alongside the flow store and survives.
- New `useResumeSnapshotHydration` hook, mounted in `MeasurementFlowContainer`. The picker's `useLoadExperimentFlow` never mounts on resume, so nothing else re-hydrates. The hook waits for the snapshots store to hydrate and re-attaches code from it when it matches the flow's `workbookVersionId`. Reports `ready`, `loading` or `unavailable`; the container shows `LoadingState` while loading. No network and no query cache involved.
- New `ResumeUnavailableState`, shown only when the flow has stripped nodes but no matching snapshots, which is reachable only if the app died between the two store writes at flow start. One message, "Can't resume this flow", and an Exit flow button that opens the existing Pause/Discard sheet. Copy in `en-US` and `nl-NL`.
- `ResolvedProtocol.code` and `ResolvedMacro.code` became optional; one narrowing in `analysis-macro-result.tsx`.
- Home "Continue" card now takes its step number, total and progress from `useFlowStepInfo`, the same hook the flow header uses, instead of counting raw `flowNodes`. It used to say "Step 9 of 9" while the header said "Step 8 of 8" because it counted the branch node. The progress bar moves from `(step-1)/total` to the header's `step/total`, so the card and header agree on both number and bar.

Persist `version` bumps to 3 and `migrate` drops anything older, the same way v1 to v2 did. A v2 flow kept its code inside `flowNodes` and has no snapshots store entry, so it would open fine on this build, lose the code from disk on its first write, and be stuck at "Can't resume this flow" on the next relaunch. Dropping it on upgrade returns the user to experiment selection instead. The cost is one paused flow lost per user who updates mid-flow, once.

## Size

Measured against the real `partialize` with a throwaway vitest probe, using the light-response-curve protocol from the IoT driver fixtures on three protocol nodes plus one 2 KB Python macro. "before" is the same state with `flowNodes` left unstripped.

| Envelope | before | after | saved |
| --- | --- | --- | --- |
| Mid-flow, with a 2 400-sample scan result | 19 113 B | 14 105 B | 5 008 B (26%) |
| Before the first scan | 7 054 B | 2 046 B | 5 008 B (71%) |

On the emulator with a real workbook (`test-vlad-3`, one PAR protocol and one JS macro), the persisted envelope was 6 881 B before the scan and 7 452 B after, with no `code` key in any of its nine nodes. The fixture protocol is under 1 KB; field protocols run 5 to 15 KB each, and the saving applies to every store write.

The code itself still lives on the device, in the snapshots store (1 160 B for `test-vlad-3`), written once when the flow loads. Total bytes at rest are roughly unchanged; the win is write frequency, once per flow instead of once per scan, step or branch visit.

## Verification

Unit: `pnpm --filter mobile test` (109 files, 1165 tests), `typecheck`, `lint` and `prettier --check` all clean. New suites: `domain/flow-snapshots.test.ts`, `hooks/__tests__/use-resume-snapshot-hydration.test.ts`. Extended: `flow-store-persistence.test.ts` pins that partialize never emits code, that a v2 envelope with code inside `flowNodes` is dropped on upgrade, that the top-level key list is unchanged, that in-memory state keeps code after `setFlowGraph`, and the snapshots store's v1 wire format.

On-device matrix, run on the Pixel 9 Pro XL emulator with the mock MultispeQ and this branch's bundle:

| # | Scenario | Result |
| --- | --- | --- |
| 1 | Resume online at the measurement node | pass, node back within 1 s, mock scan and macro ran, envelope had no code |
| 2 | Resume online at the analysis node after a scan | pass, macro result back within 1 s |
| 3 | Resume offline | pass, plus a scan afterwards |
| 4 | Resume offline with the query cache deleted | pass, resumed to the measurement node from the flow's own snapshots store, mock scan ran |
| 5 | Resume with stripped nodes and no snapshots store entry | pass, "Can't resume this flow" shown, Exit flow opens the Pause/Discard sheet |
| 6 | Questions-only flow | pass, typed answer preserved across the kill |
| 7 | Home "Continue" card | pass |
| 8 | Background/foreground only | pass, no change |
| 9 | Upgrade mid-flow: v2 envelope with code inside `flowNodes` on disk, relaunch on this build | pass, flow dropped, Home shows no in-progress card, envelope rewritten as v3 |

## Docs

- `apps/mobile/CONTEXT.md`: corrected the stale "pinned at version 0" line (measurement store is now v3, answers v1) and added one sentence on the stripped code, the snapshots store, the resume hook and the v3 drop of older flows.
- `apps/docs/content/guide/measuring/offline-sync.mdx`: one sentence that a paused flow resumes offline, plus the rare "Can't resume this flow" case.

## Linear issues

Closes OJD-1648


petar-INFO (commented):
### Via Fable

Blocking: the ticket says "preserve offline resume". Before this PR a paused flow always resumed offline. After it, resume requires the persisted getWorkbookVersion query; if it's gone and the device is offline, the user gets "Can't resume offline" until they reconnect. That is not preserved.

The cache is lost on: buster bump, storage clear, and removeOldestQuery evicting the query with the oldest dataUpdatedAt when a persist write fails. A paused flow's version query is a likely eviction target (fetched at flow start, refetched only while the flow screen is mounted on reconnect). The third case is not in the body.

Needs either evidence the eviction cases don't happen in practice, or resume that works without the cache.

---

`apps/mobile/src/features/measurement-flow/hooks/use-resume-snapshot-hydration.ts:54-59`

`error` state is sticky. With our query defaults (retry 1, refetchOnMount false, refetchOnWindowFocus false, refetchOnReconnect true) nothing refetches an errored query until an offline-to-online event. Copy says "wait a moment"; that never resolves while online. Add Retry via `refetch`, or fix the copy. Comment at resume-unavailable-state.tsx:27 is only true for `offline`.

`apps/mobile/src/features/measurement-flow/hooks/use-resume-snapshot-hydration.ts:12-15`

`errorStatus` is a cast. Existing pattern: `error instanceof ORPCError && error.status === 404` (ensure-device-registered.ts:62).

`apps/mobile/src/features/home/components/home-continue-card.tsx:24`

Unrelated fix. Also changes the bar from (step-1)/total to step/total. Not in the body.

vladstoe:
Thanks, the blocking point was right. The ticket text said "re-hydrate from the query cache" and "preserve offline resume", and those only agree while the cache survives a buster bump or an eviction, which it does not have to.

Changed approach: the flow's protocol/macro code is now written once, when the flow loads, to its own small persisted store (`measurement-flow-snapshots-storage`, v1), keyed by `workbookVersionId` and versioned alongside the flow store. Resume reads only that store. The query cache is not involved in resume anymore, so a buster bump, an eviction or being offline cannot separate a paused flow from its code. The per-write saving in the flow store is unchanged; total bytes at rest are roughly what they were, the win is write frequency.

Verified on the emulator with the query cache deleted and airplane mode on: kill, Continue, the flow resumes to the measurement node and a mock scan runs. Also re-ran the measurement-node, analysis-node and questions-only resumes against the final code.

On the smaller points: the query path is gone, so the sticky error state, the "wait a moment" copy and the `errorStatus` cast went with it.

Home card: kept. It is a one-file fix that came out of testing this change and I would rather not spin a separate PR for it. The body now spells out that the progress bar moved from `(step-1)/total` to `step/total` to match the header.

petar-INFO:
### Follow-up, again via Fable:

The query-cache blocker is resolved; the other three comments are moot or addressed.

One new blocker.

Scenario: a user is mid-flow on the current app version, pauses, updates to this build, and continues.

What happens:
- First launch on this build: the old saved flow still has its protocol/macro code inside flowNodes, so the flow opens normally. The new snapshots store is empty, because it's only written in useLoadExperimentFlow, which runs when a flow is started from the picker, not on resume.
- The user takes one step, or scans. Any write to the flow store now strips the code from the saved copy. The snapshots store is still empty.
- They pause again, or the app is killed. Next launch: the saved flow has no code and the snapshots store has nothing for it. The hook returns "unavailable". The user sees "Can't resume this flow" and has to discard.

I reproduced this with a throwaway test against the real stores. The PR body says flows paused on the previous build "never hit this path", and the hook comment says "unavailable" is only reachable via a kill between the two writes at flow start. Both are wrong for this case, and this case is every user who updates mid-flow.

Two ways to fix:
1. On a legacy resume (code present in flowNodes, snapshots store not holding this workbookVersionId), build the snapshots from the nodes' own code and write them to the snapshots store before the first strip can happen.
2. Bump the flow store to v3 and drop older flows in migrate, which is what v1 to v2 did. Simpler, but the user loses the paused flow on update.

Add the scenario to the on-device matrix either way.

---

`apps/mobile/src/features/measurement-flow/stores/use-measurement-flow-store.ts:198-200`
Comment still says the code lives in the persisted workbook-version query. It lives in the snapshots store.

`apps/mobile/src/features/measurement-flow/stores/use-flow-snapshots-store.ts:17`
"Resume reads this before the cached workbook-version query" suggests the query is a fallback. Nothing reads the query on resume anymore.

vladstoe:
**Upgrade mid-flow:** went with option 2. Flow store is persist `version: 3`, `migrate` drops anything older like v1 to v2 did. Test added for a v2 envelope with code rehydrating as empty v3, and matrix row 9 covers it on the emulator: seeded v2 flow, relaunch, no in-progress card, envelope rewritten as v3.

**Comments:** both fixed in `fe0cceaad`.

petar-INFO (approved):
pls test thoroughly with QA, as its an important touchpoint

PR #2086, opened 2026-09-14
feat(iot): replace the certificate-shaped device status with a setup ladder, and let devices be retired

petar-INFO:

- [x] Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Test Update
- [ ] Chore
- [ ] Other (please describe)

## Description

A device's status described its certificate machinery (pending, active, rotating, revoked), not where an operator stood in setting it up. This replaces it with a setup ladder that reads as progress, and adds retirement as a real end state.

The stored status becomes registered, active, revoked, retired: pending becomes registered, rotating folds into active (it only ever existed inside a single request), and retired is new. The badge never shows "active" as a word, though. It shows a rung derived from the stored status and the device's bound-experiment count, so an active device with no experiments reads as Provisioned and one with experiments reads as Onboarded. The ladder an operator sees is Registered, then Provisioned, then Onboarded, with Revoked and Retired off to the side, each rung with its own icon and tone. The device list and detail responses carry the bound-experiment count so the frontend can resolve the rung without a second call.

Retiring keeps the record and its history but takes the device out of service: broker access is cut and it drops out of every attention list and picker. Reinstating brings it back as registered, since its certificate is gone, and a phone straight to active. Phones are the exception throughout: with no certificate and their experiment chosen in the app, they have no ladder to climb, so a phone shows no status badge at all unless it has been retired, and retiring a phone cuts it off the same way.

## Linear issues

- Contributes to OJD-1768


petar-INFO:
@vladstoe @selianaaaa tldr -> changing the device status enum and how it works a bit to be clearer to the user (this was also intentionally left deferred until the largest chunks of the iot domain are implemented).. lots of touch points + migration files so lots of lines

apps/backend/src/iot/application/use-cases/list-experiment-devices/list-experiment-devices.ts
```diff
@@ -115,6 +116,14 @@ export class ListExperimentDevicesUseCase {
...
 
     const thingNames = [...boundThings, ...unboundDevices.map((device) => device.thingName)];
+    const bindingCountsResult = await this.experimentDeviceRepository.countByDevices([
+      ...bindings.map((binding) => binding.device.id),
+      ...unboundDevices.map((device) => device.id),
+    ]);
+    if (bindingCountsResult.isFailure()) {
+      return failure(bindingCountsResult.error);
+    }
+    const bindingCounts = bindingCountsResult.value;
```
selianaaaa: Smth like that would be cleaner imo:

```
const deviceIds = [
  ...bindings.map((binding) => binding.device.id),
  ...unboundDevices.map((device) => device.id),
];

const bindingCountsResult =
  await this.experimentDeviceRepository.countByDevices(deviceIds);
```

petar-INFO: applied

selianaaaa (approved):

PR #2089, opened 2026-09-15
feat(mobile): discover and join organizations from the app

vladstoe:

- [x] Feature
- [ ] Bug Fix
- [x] Refactor
- [x] Documentation Update
- [ ] Performance Improvement
- [ ] Test Update
- [ ] Chore
- [ ] Other (please describe)

## Description

Adds the requester's side of organizations to the Android app. A signed-in user can browse the organization directory with server-side search, open an organization, request to join with an optional message, see the request as pending, withdraw it, and see the member state and role once approved. A Home card nudges anyone who belongs to no organization into the directory and then tracks their request until it is decided.

Pure client work against the Phase 4 organization APIs (OJD-1718). No contract, backend or database changes. Join state is read only from the server's `membershipStatus`; there is no local optimistic state.

Also in this PR, as its own commit: the three Home cards (device, What's new, organizations) now render through one shared `HomeNavCard` instead of three copies of the same markup.

## Changes Made

- `features/organizations`: domain rules (`membership`, `join-message`, `website`), hooks for the directory, one organization, request and cancel, onboarding state and debounce; directory and detail screens; card, membership tags, type and role label maps, join sheet, three-state join CTA, open-on-web header action, unavailable state
- Routes `/organizations` and `/organizations/[id]` registered in the signed-in stack with themed headers
- Entry points: Organizations row on Profile, Home onboarding card
- `organizations` i18n namespace in `en-US` and `nl-NL` (50 keys, parity checked); orphaned `profile.organization*` keys removed
- `apps/mobile/CONTEXT.md`: Organizations glossary (directory, membership status, join request) and vocabulary line
- `eslint.config.mjs`: `organizations` added to the feature boundary list
- `features/home`: shared `HomeNavCard`; device, What's new and organizations cards render through it
- Docs: "From the mobile app" section in Find and join an organization with two screenshots; pointer from Installing the mobile app; media manifest updated

## Screenshots/Recordings

Captured on an emulator against the local seeded stack as a seed persona; these are the two images published in the docs.

| Directory | Request to join |
| --- | --- |
|  |  |

## Additional Notes

**Deliberately deferred from the ticket.** "Org results should also appear in mobile search results alongside experiments": mobile has no server-side search surface and no destination for an experiment result, so there is nothing to appear alongside. The directory's own search uses the same ranked full-text backbone. A mixed search screen on `globalSearch` is a follow-up.

## Linear issues

- Closes OJD-1733
- Contributes to OJD-1718


petar-INFO:
Via Opus
---

apps/mobile/src/features/home/screens/home-screen/home-screen.tsx:22
apps/mobile/src/features/organizations/hooks/use-organization-onboarding-state.ts:6

The Home card fetches the entire organization directory, forever, for everyone.

HomeOrganizationsCard mounts unconditionally, so useOrganizationDirectory() runs with scope "all", no search. listDirectory (organization.repository.ts:381) is unpaged and selects `description` on every row, which mobile never renders anywhere. use-organization-directory.ts:18-19 sets refetchOnMount and refetchOnWindowFocus true against an app-wide staleTime of 0, so it refires on every Home mount and every app foreground.

For a member the card returns null, so the whole payload is fetched and thrown away. That is the steady state for most users after onboarding, and it costs them bytes on a field device.

scope "related" answers "am I a member" in a small response. Only a user with zero memberships needs the wide list for the pending check, and there is no "my join requests" endpoint so that case genuinely has to ask wide. Gate it: related first, full directory only when related comes back empty.

apps/mobile/src/features/organizations/hooks/use-organization-directory.ts:16
No meta: { suppressToast: true }, unlike use-organization.ts:15. A failing list hits the global QueryCache.onError toast. On Home that is an error toast for a card the user never asked for and cannot see; on the directory screen it is a toast on top of the inline message and Retry.

apps/mobile/src/features/organizations/components/organization-membership-tag.tsx:15,22
`className` is never passed by either caller. Its default is the full layout string, so the prop can only break the layout, never help. Drop it.

apps/mobile/src/shared/ui/Button.tsx:139
Correct fix and the tests pin it, but nothing passes className to Button. Only the three cases in the new Button.test.tsx do. Unrelated shared-component change with no caller in this PR.

vladstoe:
Thanks, all four are fair and addressed:

1. Home card now asks `scope: "related"` first and only falls back to the full directory when that comes back empty, since a pending request can only show up in the wide list. Members no longer fetch the directory at all.
2. Added `suppressToast` to the directory hook.
3. Dropped the unused `className` prop on the membership tags.
4. Reverted the Button change entirely. Nothing passes a className to it, so it has no place in this PR. If a caller ever needs one it can come back in its own PR.

petar-INFO (approved):
lgtm, i guess address coderabbit comments though

PR #2096, opened 2026-09-17
feat(web,ui): readable device-by-hour heatmaps: chronological date axes, no gap interpolation, present / absent mode

Ludo-lab:
## What

Three changes to the heatmap chart, motivated by a device-by-hour "did this Ambyte report?" view on an experiment dashboard.

1. **Chronological temporal axes.** Hour buckets reach the pivot as ISO strings in GROUP BY order, and the chart-data path sends no ORDER BY, so the time axis rendered shuffled. `pivotToMatrix` now sorts an axis whose every category parses as an ISO-8601 timestamp. Plain string axes keep first-seen order. Also fixes contour, which shares the pivot.
2. **No interpolation across empty cells.** The `Heatmap` component defaulted `connectgaps` on, so Plotly invented a value for every empty cell from its neighbours and coloured it; a missing hour looked like a slightly lower reading (tooltips showed values like 12.07 on a count). The heatmap transform now sets `connectgaps: false`.
3. **Date axes and a present / absent cell mode.** The component treated ISO strings as categories, one rotated label per bucket; it now gives timestamp axes a Plotly `date` axis. A new "Cell mode" style option colours a cell green when its value is at least a threshold (default 1) and red otherwise, treating an empty cell as absent. The flat-Z and sparse-grid guards are skipped in that mode, since an all-present grid is a legitimate reading. Translations added for en-US, de-DE, nl-NL.

## Why

The Nergena barley trial dashboard needs a fleet connectivity view: 131+ Ambytes by hour, with a heartbeat every 5 minutes. A continuous count with interpolated gaps and a shuffled ISO-string axis was unreadable (screenshot in the linked conversation). Present / absent is the question the field team asks.

## Verification

- `apps/web`: 75 tests pass across the heatmap, pivot and contour suites (4 new: chronological sort, mixed-category fallback, binary mapping and threshold, all-present grid not flagged flat).
- ESLint and Prettier pass on all changed files (`packages/ui` linted with its theme config).
- `tsc --noEmit` for `apps/web` and `packages/ui` reports no errors in the touched files; the remaining errors are pre-existing and unrelated.
- No docs page describes per-chart style options, so `apps/docs/content` is unchanged.

## Notes for review

- `heatmapBinary` / `heatmapBinaryThreshold` live in the opaque `config` record, so older saved visualizations are unaffected.
- The two fixed colours (`#d62728`, `#2ca02c`) match the Databricks monitoring dashboard the field team already reads.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Ludo-lab:
Closing: the team prefers to solve this with existing dashboard capabilities rather than a code change.

PR #2101, opened 2026-09-19
feat(infra): provision the calibration sandbox Lambda and wire the backend to invoke it

petar-INFO:

- [x] Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Test Update
- [ ] Chore
- [ ] Other (please describe)

## Description

Stands the calibration sandbox up in AWS. A reusable calibration-lambda module provisions the function with its IAM, and a sandbox module wires the container through ECR into an isolated network: a dedicated security group with no egress and VPC flow logs, matching the zero-egress trust model the sandbox assumes. Both are wired into dev and prod.

The image has its own delivery path. A deploy workflow builds and scans it (with a scoped trivyignore for findings that do not apply), the production checks gate on the scan, and the release workflows carry it alongside the rest. Grafana gains a calibration dashboard row and alert rules over the new function, so a sandbox that starts failing or slowing is visible rather than silent.

## Linear issues

- Closes OJD-1755


dominikvrbic (changes requested):
## Stack review: #2101 (sandbox Lambda infrastructure)

**Verified here:** `tofu fmt -check` and backend-disabled `validate` pass for dev and prod.

### P0, blocks deploy

1. **First apply cannot create the Lambda.** `infrastructure/modules/calibration-lambda/main.tf:160` points the function at `calibration-sandbox-<env>:latest`. The same apply creates an empty ECR repository, and `.github/workflows/deploy-calibration-sandbox.yml:82-92` pushes only a commit-SHA tag, downstream of the infrastructure job. Neither side can go first. Needs a bootstrap image, a two-phase apply, or a `latest`-tag push in the workflow with the Lambda created after it.

2. **Backend task omits the endpoint variable and crashes at boot.** `infrastructure/env/dev/main.tf:2238-2241` and `infrastructure/env/prod/main.tf:2198-2201` inject `AWS_LAMBDA_CALIBRATION_SANDBOX_FUNCTION_NAME` but never `AWS_LAMBDA_CALIBRATION_SANDBOX_ENDPOINT`, which `config.service.ts:67-69` (from #2035) loads with `getOrThrow`. No ECS secret, SSM parameter, or other task-definition path supplies it. `infrastructure/modules/calibration-sandbox/README.md:53` claims the backend degrades gracefully; the code contradicts it. The right fix is on the backend side (#2035, make it optional); until then the task defs need the variable injected as empty.

3. **"Zero egress" allows HTTPS to the whole VPC.** `infrastructure/modules/vpc/main.tf:344-353` permits TCP 443 to the full VPC CIDR rather than the three interface endpoints. The comment at 333-335 assumes the function ENI needs ECR for cold starts, but Lambda pulls the image as a service operation and ships logs via the execution role, not the VPC route. Remove the rule. Also remove the redundant execution-role ECR allow at `calibration-lambda/main.tf:63-85` (the repository policy already grants the Lambda service principal); as written it hands escaped code ECR credentials. This matters more than usual because #2035's in-process isolation does not hold; this layer is the actual boundary.

4. **Function code retains ENI management permissions.** `calibration-lambda/main.tf:42-60` grants ENI create/delete/assign to the execution role, and the deny at 128-131 only covers `RunInstances`, `TerminateInstances`, `ModifyInstanceAttribute`. Add the AWS-recommended conditional deny on `lambda:SourceFunctionArn` so function code cannot use them while the Lambda service still can.

5. **Release deploy does not scan the image it ships.** `deploy-calibration-sandbox.yml:76-106` builds, pushes and installs without Trivy; the scan in `production-checks.yml:66-78` is a separate PR-time job. A release from an arbitrary ref or a base-image update deploys an unscanned image.

6. **Immutable SHA tags break reruns.** The workflow always pushes the checked-out SHA; dev sets `IMMUTABLE` (`env/dev/main.tf:326-328`) and prod inherits it. Any rerun gets `ImageTagAlreadyExistsException`.

7. **Duplicate flow logs, misattributed metric.** `calibration-sandbox/main.tf:25-33` creates `ALL` flow logs for the isolated subnets that `macro-sandbox/main.tf:28-36` already logs, doubling cost. The REJECT metric filter at `calibration-lambda/main.tf:220-229` matches every reject in the copy without filtering by this function's ENIs.

### P2

- `grafana/dashboard/main.tf:1100-1176`: the error alert needs `> 2` for a sustained 5 minutes, and the throttle rule (1188-1251) has the same hold. One failed bench session or one throttle ages out before either fires. Lower the threshold and hold for a low-volume function.

petar-INFO:
All seven are fixed, and the P2. `tofu fmt -recursive` is clean and both environments validate.

### P0

**1. First apply.** The workflow no longer assumes an image is already there. It checks whether the commit's image exists in ECR, builds only if not, scans, then pushes the SHA tag and, when the repository has no tags at all, the `latest` tag alongside it. Updating the function code is skipped with a recorded `updated=false` when the function does not exist yet, so the first apply can create it against a tag that is now really there. The module README documents the two-step bootstrap rather than leaving it to be discovered.

**2. Endpoint variable.** Fixed on the backend, as you suggested, in #2035: it is read with `get(...) ?? ""` instead of `getOrThrow`, with a test that boots with the variable absent. The task definitions do not need it injected.

### P1

**3. Egress.** Both sandbox egress rules now target the endpoints' security group by `source_security_group_id` rather than the VPC CIDR, and the execution role's ECR allow is gone, so escaped code has no ECR credentials to use. Worth saying explicitly: this matters more now that #2035's hook exists, not less. The hook is a layer; this is still the boundary.

**4. ENI management.** Added the conditional deny on Create and Delete NetworkInterface and Assign and Unassign PrivateIpAddresses, with `Condition = { Null = { "lambda:SourceFunctionArn" = "false" } }`, so the Lambda service can still attach the ENI while function code cannot.

**5. Scanning.** A Trivy step (pinned by digest) runs between the build and the push, with per-app ignore files, so nothing ships unscanned from a release or a base-image bump.

**6. Immutable tags.** The workflow checks for the tag first and skips the build and the push when the SHA is already in the repository, so a rerun no longer dies on `ImageTagAlreadyExistsException`.

**7. Flow logs.** The macro sandbox module now exports its flow log group name and the calibration module takes it as a variable instead of creating a second `ALL` log for the same subnets. `flow_log_retention_days` is gone from the calibration module and from the prod environment.

### P2

**Grafana.** The error rule fires on `> 0` held for 1 minute and the throttle rule holds for 1 minute, so a single failed bench session on a low-volume function is visible instead of ageing out.

The same fixes were applied to the macro sandbox's deploy workflow and Lambda module, since the two are copies of each other and the findings applied to both.

### Where it landed

- [`37b44b888`](https://github.com/Jan-IngenHousz-Institute/open-jii/commit/37b44b888) scope sandbox egress to the endpoints, deny ENI management to function code, drop the role's ECR grant, share the flow logs, and make the sandbox deploys scan, bootstrap and rerun cleanly

The whole stack is rebased onto `main` (`8be3155b3`) and force-pushed, so every branch carries the seven commits main gained this week, and the generated API spec is regenerated on each branch that changes the contract so the drift gate passes per PR. Workspace gate on the stack head: web 6,757 tests across 947 files, backend 3,498 across 299, `@repo/api` 1,930, `@repo/iot` 604, plus lint, typecheck and the production web build. CI on this PR: 15 checks green.

dominikvrbic (approved):

PR #2102, opened 2026-09-19
feat(iot): bench instrument layer with Kiprim, CaliTool, PAR references, and Ambit console support

petar-INFO:

- [x] Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Test Update
- [ ] Chore
- [ ] Other (please describe)

## Description

Adds a bench-instrument layer to `@repo/iot`, kept deliberately separate from device drivers: an instrument is rig equipment a procedure drives but never ingests from, so nothing here is a sensor family, registered, or bound to an experiment. One interface describes what a procedure needs from a piece of equipment, its setpoints and optional readings, an identity handshake, and a shutdown that returns it to a state safe to walk away from (a lamp left at current after an aborted sweep is a hazard, not a detail).

Concrete support arrives for the Kiprim DC source, the CaliTool spectral board, and the MiniPAR and MicroPython PAR reference sensors, each recognised over serial by a substring of its identity reply. A registry probes a port through the candidates in a fixed order, with the MicroPython prompt last because its control-byte handshake would disturb an instrument that does not speak it. Alongside, the Ambit driver gains the console commands, response parsers, and boot-dump device-info parsing a rig needs from the device under test.

## Linear issues

- Contributes to OJD-1616
- Contributes to OJD-1608
- Contributes to OJD-1617
- Contributes to OJD-1613
- Contributes to OJD-1612
- Contributes to OJD-1601


dominikvrbic (changes requested):
## Stack review: #2102 (bench instrument layer)

**Verified here:** `@repo/iot` typechecks and its suite passes (153 tests at this PR, 594 at stack head). The `strictTimeout: true` path leaves the RX buffer intact on rejection, but `sendAndCollect` clears `rxBuffer` before every send, so no stale partial trace leaks into the next command. Of the seven PRs this one is the cleanest, but all three P1s are failures the fixtures cannot see and real hardware will.

1. **Boot-dump line budget counts the wrong lines.** `driver/ambit/device-info.ts:36,201`. `AMBIT_BOOT_DUMP_MAX_LINES = 26` is documented as "lines the firmware prints before the terminating FW: line" but is applied to the whole reboot reply, which (per `commands.ts`'s own REBOOT comment) starts with the ESP32 ROM + IDF boot log of roughly 30 to 45 lines. On real hardware the loop exhausts inside the bootloader log, `readDeviceInfo()` returns null, identity has no MAC, and #2103's write-back read-back has nothing to verify against. Longest fixture is 9 lines. Start the budget at the first recognised firmware line.

2. **MicroPython reader does not skip the leading blank line.** `instrument/micropython-par/instrument.ts:104-128`. `takeLines` treats the first line as the answer, unlike the sibling CaliTool reader which strips leading `\r\n` and comments on why. MicroPython emits a line break before the raw-REPL banner, so `identify()` returns `""` and `identifyBenchInstrument` reports nothing plugged in for a board that answered; `read()`'s `[, valueLine]` destructure shifts the same way. Kiprim's `awaitLine` has the same first-line assumption (an echoed `*IDN?` would be taken as the identity). Tests hard-code banners without the leading break.

3. **CaliTool `read()` gives the caller's timeout to the wrong command.** `instrument/calitool/instrument.ts:144-158`. The caller's `timeoutMs` goes to the cheap `get N` while the slow `measure` acknowledgment keeps the hardcoded 2 s default. The interpreter passes that timeout in specifically so a slow instrument can be waited out (`interpreter.ts:491`). With atime/astep near the declared ceilings the integration runs past 2 s, the read throws, and the sweep aborts.

### P2

- `instrument/calitool/instrument.ts:144-158`: each channel read re-measures, so 12 channels are 12 integrations rather than one spectrum.
- `driver/ambit/commands.ts:99-103`: the baseline completion predicate doubles as the parser; a preamble line of six integers wins.
- `instrument/micropython-par/instrument.ts:52-61`: soft reboot on identify with no settle, so the first read races the boot output.
- `instrument/kiprim/instrument.ts:77-85`: zeroing the lamp is unverified and the output is left on, unlike CaliTool's shutdown.
- `core/identify-device.ts:187-190`: a dev build reports as the released version and passes `minFirmwareVersion`.
- Reuse: three hand-rolled line readers and no command queue in `instrument/*`, beside the existing `collectReply` and `CommandQueue`.

petar-INFO:
The three P1s were all real and all invisible to the fixtures, exactly as you said. Each is now pinned by a test that fails against the old code. 468 tests at this PR, 604 at the stack head.

### P1

**1. Boot-dump budget.** The budget is no longer spent on the bootloader. `readDeviceInfo` counts only from the first line it recognises as firmware output, so the ESP32 ROM and IDF log ahead of it cost nothing. Pinned with a fixture that puts a bootloader log in front of the dump, which the old loop exhausted before reaching the MAC.

**2. Leading blank line.** The MicroPython reader now strips leading `\r\n` before taking lines, like the CaliTool reader does, so a board that answers is no longer reported as nothing plugged in, and `read()`'s destructure no longer shifts by one. Kiprim's `awaitLine` got the same treatment. The tests now carry the leading break the boards actually emit.

**3. CaliTool timeout.** The caller's `timeoutMs` now covers `expectAck("measure", ...)`, which is the slow half, and `get N` uses the instrument's own reply timeout. A slow integration near the declared atime and astep ceilings is waited out instead of aborting the sweep. There is a test at a long integration that fails against the old ordering.

**P2, soft reboot.** `identify()` now waits out the reboot banner with a quiet window before returning, so the first read no longer races the boot output.

### Not changing, with reasons

**P2, re-measuring per channel.** This is what the procedure asks for rather than a driver decision: each read step names one channel, and the driver honours it. Reading a whole spectrum once and distributing it across a step's columns is a change to the procedure grammar (a read that yields several columns from one integration), not to this class. Worth doing, and it belongs with the DSL.

**P2, baseline preamble.** Checked against the firmware source: the baseline command prints the six integers and nothing before them, so a preamble line of six integers is not a shape the device can produce. The completion predicate doubling as the parser is fine for the input that actually arrives. If that ever changes, the fix is a framing terminator, which the other long commands already use.

**P2, Kiprim output left on.** This matches the bench tool: it sets voltage and current and never touches the output switch, so an operator's supply stays in the state they left it in between runs. Zeroing the current is the safety action and it is verified by the rest sequence. Turning the output off as well would be a change in behaviour against a tool people already use, and I would rather not make it silently.

**P2, dev build reporting as a release version.** Acknowledged, and it is a real gap in the firmware gate. It is not reachable from this stack's code path and the fix belongs with whoever owns the version string on the device side.

**P2, reuse.** Fair. The three readers and the absent command queue are worth consolidating onto `collectReply` and `CommandQueue`, and that is a refactor with the whole suite as its gate rather than something to fold into a review round on new code.

### Where it landed

- [`67b7d39f7`](https://github.com/Jan-IngenHousz-Institute/open-jii/commit/67b7d39f7) count the Ambit boot dump from its first line, skip the REPL's leading break and wait out its reboot, and give the tool board's integration the caller's timeout

The whole stack is rebased onto `main` (`8be3155b3`) and force-pushed, so every branch carries the seven commits main gained this week, and the generated API spec is regenerated on each branch that changes the contract so the drift gate passes per PR. Workspace gate on the stack head: web 6,757 tests across 947 files, backend 3,498 across 299, `@repo/api` 1,930, `@repo/iot` 604, plus lint, typecheck and the production web build. CI on this PR: 17 checks green.

dominikvrbic (approved):

PR #2103, opened 2026-09-19
feat(iot): capture-procedure interpreter with device setpoints and calibration write-back

petar-INFO:

- [x] Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Test Update
- [ ] Chore
- [ ] Other (please describe)

## Description

The client runtime that executes a capture procedure against a connected rig. The interpreter is the one place a declared step becomes serial traffic: it walks the operator, settle, read, and sweep steps, driving instruments and the device through their setpoints and collecting each series, while fitting stays the sandbox's job and writing back the wizard's. A rig is bound by declared role, and a role may read, apply setpoints, or both, with an operator port for the prompts and confirmations a person has to answer at the bench.

The calibration layer is how approved coefficients reach a device, per family: an output-schema key becomes a console command, every write is checked against the device's reply, and a block the device can read back is checked against what it holds afterwards. A capability map derived from the driver, instrument, and writer registries says what a definition may actually declare, which matters most for the one failure that is otherwise silent, a coefficient block no writer covers computes, passes review, is approved, and then never reaches the device.

## Linear issues

- Contributes to OJD-1616
- Contributes to OJD-1608
- Contributes to OJD-1617
- Contributes to OJD-1613
- Contributes to OJD-1612
- Contributes to OJD-1601


dominikvrbic (changes requested):
## Stack review: #2103 (capture-procedure interpreter + write-back)

**Verified here:** `@repo/iot` typechecks clean; 594 tests across 34 files pass. Step ordering, settle timing, named columns and the reserved stimulus column all match the #2032 schema.

### P1, hardware safety

1. **No AbortSignal; a running sweep cannot be stopped.** `interpreter.ts:44-52`. The wizard's only exit (unmount) rests instruments and closes ports while the sweep loop is still stepping, so a lamp or DC source can be re-driven after the zero write and before the close, and the second rest fails on a closed port and is swallowed. Thread an `AbortSignal` through `runProcedure`, check it between setpoints and inside settle waits, and make abort itself rest the rig before resolving.

2. **Blocks are half-written when one coefficient has no writer.** `write-back.ts:383-394`. `writeBlock` sends coefficients as it iterates, so a block with one uncovered coefficient is partially on the device before "No writer for coefficient" returns. The wizard passes `applied.blocks` whole and `canWrite` needs only one writable block, so the path is live. Contradicts the "never sent" doc on `writableCalibrationBlocks`; the spec at :452 covers only a one-coefficient block. Resolve every writer for the block before sending anything.

3. **Kiprim shutdown failures are swallowed.** `shutdownRig` catches and drops the error, so a DC source that could not be zeroed is reported nowhere. Surface it to the caller as a distinct "rig not safe" outcome.

### P1, contract drift

4. **Structured device replies overflow the cell cap.** `interpreter.ts:449-463`. `toCell` JSON-stringifies structured DUT replies; MultispeQ/MiniPAR envelopes with `data_raw` are tens of KB, and `zSeriesCell` caps strings at 4096 (`iot-calibration.schema.ts:194-200`). oRPC validates before the use case, so the 400 lands after the whole bench session, and the wizard only holds readings in React state so retry re-runs the bench. Protocol reads on the DUT are the contract's normal case. Either project the reply to the declared numeric columns here, or raise the cap in #2032 for structured cells; the current combination is unrecoverable.

5. **Capability map is enforced nowhere.** `familyCalibrationCapabilities` and `benchInstrumentSummaries` have zero call sites outside their specs; the only consumer of the map is `canWriteCalibration` in the wizard, after approval. The backend stores `outputSchema` unexamined, so the silent failure this PR calls out (a block computes, passes review, is approved, and never reaches the device) is still only discovered at the write step after a full session. Also `WritableCoefficient` exposes only `{name, isArray}`, so a length mismatch against the Ambit baseline vector (exactly 6 entries) is uncatchable even client-side. Expose the map through the API and validate `outputSchema` against it at definition creation.

6. **`CaptureResult.skipped` has no contract field.** Skipped optional steps are surfaced live but dropped by `toRunPayload`, so the record has no trace of what the bench could not do.

### P2

- `interpreter.ts:239`: a sweep commits only when complete; a fault at point 9 of 10 discards 8 confirmed rows.
- `interpreter.ts:163, 213, 353-356`: an operator decline inside a read or sweep step surfaces as `ProcedureRigError`, not `ProcedureDeclined`.
- `write-back.ts:142-146, 149-152, 280`, `device-setpoints.ts:27, 44`: the firmware reply shapes the verifiers assume could not be confirmed from the repo and one is brittle; needs a bench check before merge.
- `interpreter.ts:372-375` with contract `:203-207`: `_retaken` doubling plus the 20-series payload cap means 11 retaken read steps is a 400 (bound fix belongs in #2032).

petar-INFO:
Everything here is addressed. The hardware-safety three are the important ones and they were right. 604 tests in the package.

### P1, hardware safety

**1. AbortSignal.** `ProcedureContext` takes an optional `signal`. The runner checks it before each step, before each setpoint and before each read, and a settle is now an abortable pause rather than a sleep the rig drives through, so a stop cannot land a setpoint after the zero write. Aborting raises `ProcedureStopped`, which carries no partial-run blame with it.

The browser half is on #2104: there is a Stop control on the capture step and on the check, and leaving now stops the run, waits for the interpreter to let go, rests and closes the bench, and only then releases the device driver. The connection hook and the rig hook each grew an opt-out from tearing themselves down on unmount so that one owner can order the whole sequence, rather than React's hook order deciding it.

**2. Half-written blocks.** `writeBlock` now resolves every coefficient's writer into a list before it sends anything, so a block with one uncovered coefficient is refused whole and nothing reaches the device. A test covers a two-coefficient block with one writer, which is the case the old spec missed.

**3. Shutdown failures.** `shutdownRig` attempts every instrument and then throws with each failure named: "The rig is not at rest. kiprim-dc: port closed". The rig hook aggregates the device's own rest into the same error, and the wizard shows it to the operator instead of logging it. A supply that would not zero is now something a person hears about.

### P1, contract drift

**4. Cell cap.** `MAX_SERIES_CELL_TEXT` is 65,536, held by value on both sides (the device package cannot import the contract), and `readOnce` refuses a longer cell at the read. The operator hears about it at the step that produced it, with the column named, instead of losing the session to a 400 at submit. The contract's cap was raised to match in #2032.

**6. Skipped steps.** `CaptureResult.skipped` now has somewhere to go: `zSkippedSeriesList` on the contract, a `skipped_series` column with a migration, the wizard sends it with the run, and the run record shows what the bench could not do and why.

**P2, partial sweeps.** A sweep commits the rows it confirmed before the abort propagates, so a fault at point 9 of 10 keeps the 8 that were good.

**P2, declines.** An operator declining inside a read or a sweep now raises `ProcedureDeclined` with the prompt, not `ProcedureRigError`.

**P2, `_retaken` bound.** Covered by the bound reconciliation in #2032: the payload series cap is now derived from the step cap with room for a retake on every step.

**P2, verifier reply shapes.** Still flagged on the file for the desk run. They cannot be confirmed from the repo and I would rather they stay marked as unconfirmed than be quietly treated as known.

### Partly taking, with a note

**5. Capability map.** Two halves here, and I agree with one of them.

`WritableCoefficient` now exposes `length` where the writer knows it, populated from the vector writer's entry count, so the Ambit baseline's six entries are declarable and the authoring UI fills that length in rather than guessing ten. That closes the uncatchable mismatch.

Validating `outputSchema` against the map at definition creation is the half I am not taking. The map says what the platform can write today, and a definition is allowed to compute more than that: a block that is approved on record and never written is a supported outcome, which is the only thing available for families with no writers at all. Enforcing the map at creation would make the device package's current coverage a hard limit on what a bench may measure, and would invalidate stored definitions whenever a writer is added or renamed. What the author needs is to know, not to be stopped, so the authoring surface offers the writable block and coefficient names directly and says plainly which declared coefficients have no writer. That is in the library branch, and it moves the discovery from the write step to the moment of writing the definition, which is the actual complaint.

### Where it landed

- [`a42eff2d8`](https://github.com/Jan-IngenHousz-Institute/open-jii/commit/a42eff2d8) let a run be stopped between what it does, keep a cut sweep's rows, refuse an oversized reply at the read, resolve every writer before sending a block, and report a rig that would not rest

The whole stack is rebased onto `main` (`8be3155b3`) and force-pushed, so every branch carries the seven commits main gained this week, and the generated API spec is regenerated on each branch that changes the contract so the drift gate passes per PR. Workspace gate on the stack head: web 6,757 tests across 947 files, backend 3,498 across 299, `@repo/api` 1,930, `@repo/iot` 604, plus lint, typecheck and the production web build. CI on this PR: 17 checks green.

dominikvrbic (approved):

PR #2104, opened 2026-09-19
feat(web): device calibration tab with the bench run wizard and review

petar-INFO:

- [x] Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Test Update
- [ ] Chore
- [ ] Other (please describe)

## Description

The device Calibration tab, where a run happens end to end. A rig hook opens the serial ports, identifies each connected instrument by its handshake, and binds it to the declared role, tracking each role as connected, mismatched, unrecognised, or missing, and telling a required role apart from an optional step's. The wizard then walks the operator through connecting the rig, picking the definition, and the capture itself, driving the interpreter's sweep while surfacing the operator prompts and reading confirmations the procedure calls for.

Once captured, the run is recorded and reviewed: its coefficient blocks, fit chart, and series table are shown, and a reviewer approves or rejects it. An approved calibration can be written back to the hardware over the same connection and its write recorded against the applied row, and the active calibration is shown on the device. Seeded bench procedures make the whole flow runnable against the local stack.

## Linear issues

- Contributes to OJD-1616
- Contributes to OJD-1608
- Contributes to OJD-1617
- Contributes to OJD-1613
- Contributes to OJD-1612
- Contributes to OJD-1601


dominikvrbic (changes requested):
## Stack review: #2104 (device Calibration tab + bench wizard)

Part of a coordinated review of the 7-PR calibration stack (#2032 → #2104) run as ten independent reviewer passes (Fable, Opus, GPT-5.6 Sol, GPT-6 Astra, Kimi K3), each validating its findings against the code before reporting. Two reviewers covered this PR independently (design/product and file-by-file correctness). Stack head reviewed: `54d3d8ed1`.

**Verified here:** web typecheck and eslint pass; calibration suites (62 files, 296 tests), ui wizard and data-table suites pass; all five seeds parse through `zCreateCalibrationDefinitionBody`; the seed split is faithful; Web Serial `close()` cannot race a reader lock because the adapter releases locks first.

### Blocker-adjacent

1. **Coefficients are written to whichever unit is on the port.** `calibration-connect-step.tsx:56-59`, `useCalibrationCapture.ts:97`. The only gate is a sensor-family check; nothing compares the identified `DeviceIdentity` (name/deviceId; `iot_devices.serial_number` is unique) to the route's `deviceId`. Two MiniPARs on a bench: approve for A, write onto B, the record says A. Bind the identified device to the route device before capture and again before write, and send the identity with the write report so the backend can refuse a mismatch.

### P1, hardware safety

2. **Unmount is the only mid-run teardown, and it tears down in the wrong order.** `useCalibrationCapture.ts:42-49`, `useIotConnections.ts:52-58`, `useCalibrationRig.ts:255-299`, `calibration-wizard.tsx:410`. There is no Stop control. On navigation, cleanup declaration order destroys the DUT driver before the rig's `rest()` runs, so the device under test is left driving its LEDs; meanwhile the lamp/DC-source rest races the still-running sweep (the interpreter has no AbortSignal, see #2103) so a setpoint can land after the zero write. Both rest failures are swallowed. Needs a Stop button wired to an abort that rests the rig first, and cleanup ordered rig-then-device.

### P1, unrecoverable sessions

3. **A failed run POST forces a full re-run of the bench.** `calibration-wizard.tsx:149-171, 181-183`. On any `createCalibrationRun` failure, including the deterministic 400 from oversized structured cells (#2103 finding 4), the captured payload in state is never reused; retry re-runs the sweep and fails identically. Keep the payload and retry the POST alone.

4. **Completed sweep dropped silently.** `calibration-wizard.tsx:151`, `useCalibrationCapture.ts:61`. When the DUT connection object is gone while the writer still works, the capture is discarded and the step renders no buttons.

5. **Preset-definition path dead-ends at capture.** `calibration-wizard.tsx:96-99, 374, 410-422`. With `procedure` undefined, `hasEveryRequiredRole` is vacuously true, `capture()` returns null silently, `captureStartedRef` blocks any re-fire, and nothing renders.

6. **Write step has no recovery.** `calibration-wizard.tsx:216, 232-243`. It says "reconnect and write again" with no reconnect control and no later path to write an approved-but-unwritten calibration. A device write whose report POST fails gets no retry while Done says "confirmed", so hardware and record diverge.

7. **Approve/reject exist only inside the live wizard.** Run detail has no decision controls, so a computed run is stuck after a reload.

8. **Firmware suffixes silently dropped.** The wizard's `zFirmwareVersion` gate discards versions with a suffix, so a floored definition rejects with a false "device did not report a version".

9. **Definition links 404.** Search, org resource rows and sharing routes register `calibration_definition` under `calibrations`, but no `/platform/calibrations/[id]` page exists (see #2033 finding 3).

### P2

- `toRunPayload` drops `CaptureResult.skipped`; skipped optional steps leave no trace in the record.
- `clear.seed.ts` now deletes every experiment in the seed user's personal org; deliberate, but deserves a loud comment. Re-seeding after the wizard has been used as the seed user hits an FK violation.
- Tab-level and page-level permission gating disagree; series tables are unpaged.
- No `apps/docs/content` for the Calibration tab, six-step wizard, operator prompts, review gate or write-back. AGENTS.md makes docs a same-PR requirement.

petar-INFO:
This was the most substantial of the seven and nearly all of it was right. Everything below is fixed on this branch, with tests, and the figures at the end are the gate after the whole stack was rebased onto main.

### Blocker-adjacent

**1. Coefficients written to whichever unit is on the port.** The wizard now takes the device's registered `serialNumber` and compares it with the identifier the unit announces, through `serialsMatch` from the contract, which compares on alphanumerics only so a MAC printed with separators still matches one printed without. There are three outcomes and the connect step shows each one: a match confirms the serial on the device row, a mismatch names both serials and disables Continue, and a family that announces no identifier at all says so plainly rather than letting silence read as a check that passed. MiniPAR is that third case, which is why the contract field is optional. The serial travels with the run and with the write report, and #2033 refuses a mismatch server-side as well.

### P1, hardware safety

**2. Stop, and teardown order.** There is a Stop control on the capture step and on the post-write check, wired to an `AbortController` threaded into the interpreter (#2103). Stopping also cancels whatever prompt the operator was being asked, so the run unwinds rather than sitting on a promise.

Leaving is now one ordered sequence owned by the capture hook: stop, wait for the interpreter to let go, rest and close the bench, then release the device driver. To get that order, `useIotConnections` and `useCalibrationRig` each take an opt-out from destroying themselves on unmount, so React's hook order no longer decides whether the device's driver outlives the rest that has to go through it. `shutdownAll` is idempotent, since close, unmount and a change of procedure can all ask for it.

Rest failures are no longer swallowed on either side: the rig aggregates the device's and the bench's into one error and the operator sees it.

### P1, unrecoverable sessions

**3. A failed POST no longer costs the bench session.** Capture and submission are separate. The readings and the skipped list stay in state, the failure says so, and the button offers to record those readings again rather than to run the sweep again. There is a test that fails the POST, retries it, and asserts the device was read three times in total rather than six.

**4. A completed sweep is no longer dropped.** A missing connection is a named outcome rather than a silent `return`, so the step renders the reason and a way forward instead of nothing at all.

**5. Preset path.** `canStart` now requires a loaded procedure, so the preset-definition route cannot walk into a capture step that has nothing to run.

**6. Write recovery.** The write and the recording of it are separate outcomes with separate errors. A failed report offers to record the write again without touching the device, since the coefficients are already on it and writing twice would be the wrong repair. A port pulled between approving and writing offers to reconnect rather than leaving an enabled button that cannot work. Done is truthful about all of this: it distinguishes confirmed, written but not confirmed, written but not recorded, approved on record with nothing written, and a family the platform cannot write to at all.

**7. Decisions outside the live wizard.** The run detail carries them now: approve and reject when the run is still computed, and a "Write to device" action when a calibration was approved but never reached the hardware, which opens the wizard in a write-only session of connect, write, done. A run left computed when a tab closed is no longer stuck.

**8. Firmware suffixes.** Fixed at the contract, in #2032. There is a second schema for what a device reports, which accepts a release suffix, and the wizard sends it instead of dropping it. A backend test pins that such a version passes a definition's floor.

### P2

- `skipped` now travels: the wizard sends it with the run and the run record lists each skipped series with its reason.
- `clear.seed.ts`: the FK violation is fixed properly rather than commented. The clear now also collects definitions created by a seed user or living in a seed user's personal organization, deletes their runs and their grants, and only then the definitions, so a second `db:seed` after using the wizard as the seed user no longer trips over a definition the platform itself created.

### Not changing

**9. Definition links.** The route exists on `feat/calibration-library`, the eighth branch, whose PR is not open yet. The page and the segment registration land together.

**P2, gating and paging.** Noted. The tab-level and page-level checks disagree in the direction of showing less rather than more, and the run list is bounded by a device's own history. Both are worth a pass, neither is a defect in this diff.

**P2, docs.** The calibration documentation is being written once at the end of the programme rather than accreting across eight intermediate PRs that redefine the same surfaces as they go. That is a deliberate decision for this stack.

### Where it landed

- [`eff1887bb`](https://github.com/Jan-IngenHousz-Institute/open-jii/commit/eff1887bb) bind a session to the unit that answers, let a run be stopped, and give a failed submission, write or report its own way back
- [`0aa049c84`](https://github.com/Jan-IngenHousz-Institute/open-jii/commit/0aa049c84) give the seeded Ambit the serial its console reports, so the seeded bench can run
- [`f2e6cfc69`](https://github.com/Jan-IngenHousz-Institute/open-jii/commit/f2e6cfc69) load the calibration charts client-side only, so Plotly stays out of the server bundle

dominikvrbic (approved):

PR #2105, opened 2026-09-19
perf(data): reduce macro execution task size

dominikvrbic:
## Summary

- range-partition exploded macro rows into 128 tasks by `(macro_id, coalesced workbook_version_id, id)` immediately before the pandas UDF, preserving source, flow, table, pipeline, and checkpoint identities
- execute strictly sequential, homogeneous `(macro_id, workbook_version_id)` chunks of at most 25 items; preserve response order and duplicate multiplicity with per-key FIFO queues
- charge a full six-second delay before every macro POST, including a new task/client's first request; add no Python POST retries

## Why range partitioning

Read-only production SQL showed that round-robin distribution scatters each macro/version group across most partitions and inflates homogeneous POST count on ordinary batches:

| Rows | Round-robin homogeneous POSTs | Range homogeneous POSTs |
| ---: | ---: | ---: |
| 2,023 | 334 | 130 |
| 10,705 | 665 | 512 |
| 184,358 | 7,572 | 7,440 |
| 324,653 | 13,952 | 13,049 |

The exact range candidate was also simulated on the current 324,653-row production workload, where one macro/version group held about 80% of rows. It populated all 128 partitions with 2,188-3,001 rows (p95 2,818), created only 137 group fragments, and added 62 POSTs over the no-fragmentation ideal (0.5%). Pair-only hash partitioning put 80% of rows in one task; including high-cardinality `id` in the range key is load-bearing.

These are measured fixtures, not a guarantee: boundaries are sampled, small batches need not populate every partition, and equal full keys can remain skewed. Local regressions cover those limits.

## Conditional request-start bound

This is a Spark-conditional request-start bound, not a distributed rate limiter. Production is now verified at four workers with two task slots each and driver `maxTasks = 0`, under policy range 1-4. While that ceiling remains true, the full pre-request charge limits macro POST starts to:

`8 * (floor(300 / 6) + 1) = 408` per rolling five-minute window.

Raising the worker-policy ceiling invalidates the bound. It does not cover non-Spark callers sharing the WAF key or backend work that outlives a failed/speculated task.

## Fan-out and retry semantics

Each POST contains one macro/version group. The backend creates the same groups and makes one logical `lambdaPort.invokeLambda` call for a valid group. This prevents cross-group fan-out, but it does not prove one physical execution because the AWS SDK client has `maxAttempts: 5`; macro execution is not exactly-once today.

No status-only retry was added. The endpoint has no idempotency key, a lost response is ambiguous, and macros may be nondeterministic. A generic 403/429 does not prove an edge rejection. Transport/status failures remain explicit ordered failed rows; HTTP-200 partial failures are not retried.

## Validation

- `uv run pytest -q` — 117 passed
- `uv run ruff check .` and `uv run ruff format --check .` — passed
- `uv run pyright` — 0 errors, 12 existing warnings
- tests cover range fragmentation across repeated sampled runs, null workbook versions, small batches, equal-key skew, exact cardinality, homogeneous chunking, duplicate multiplicity, reversed backend response order, response-cardinality mismatch logging, full delay before first/every POST, and one call for 429, 503, transport failure, and HTTP-200 partial failure
- an actual parquet `readStream` is passed through the range partition helper and executed through an available-now memory-sink trigger, preserving all 96 rows and association fields; range partitioning is therefore exercised through Spark's streaming analyzer and execution path, not only batch DataFrames

The streaming test uses local Spark rather than DBR/DLT and an available-now trigger rather than production's continuous 120-second trigger. Production scheduling and sampling overhead remain canary measurements.

## Limits and canary gates

Fixed 128 is task-size reduction, not a row/time bound. Task duration still grows with admitted micro-batch volume; source-side admission control remains a separate future design. A large duplicate full key can still collapse into one task, so maximum task rows is a mandatory canary metric.

The four-worker/eight-task premise is satisfied. The first live observation at that size still had four macro slots plus four enrichment slots consume all capacity while Kinesis Buffering waited and raw-data lag reached 26 minutes. Do not deploy until WAF aggregation/non-macro shared-key traffic and Lambda/API throttle and latency metrics are read.

Canary with a normal checkpoint-resuming update and include the daily approximately 05:15-06:10 UTC burst. Measure range-sampling overhead, actual partition distribution and maximum task rows, task wall time, shuffle/spill/GC, POST starts, backend 429/5xx/latency, Kinesis Buffering queue wait, and row/error cardinality. Rollback is a normal checkpoint-resuming code revert only. Do not reset/move checkpoints, full-refresh, replay Kinesis, change the source, or change capacity as part of rollback.

## Linear issues

Contributes to OJD-1836.

## Additional notes

This draft performs no live deployment, restart, replay, scale change, merge, or checkpoint operation. It is intentionally separate from PR #2100.


petar-INFO:
Read this while working on the pipeline overhaul (OJD-1958), which touches the same chunker, so treat it as findings to check rather than review comments. The measured fixtures and the stated limits in the description made it much easier to check than most.

The group-strict chunking, the FIFO queues for duplicate macro keys, and the cardinality mismatch logging are all clearly right, and are three things my branch was missing. Declining to add retries is well argued too. Three findings below, and the first two are about the six second delay.

## The sleep is charged against a Spark slot

`time.sleep` inside the pandas UDF holds an executor slot while doing nothing. On the 324,653 row workload in your table that is 13,049 requests times six seconds, so roughly 78,000 slot seconds, or about 2.7 hours with eight slots held idle.

Slot contention is the thing the range partitioning is trying to relieve. The September figures have the Kinesis prefetch stage doing 15 seconds of work while waiting a p90 of 752 seconds for a slot, and 5,777 seconds of queueing against 495 seconds of work. A blocking delay in the UDF adds to that same queue, which is why I think this one matters more than the arithmetic below.

If the request rate genuinely needs bounding, fewer and larger requests bound it without holding a slot, and a backend side 429 with `Retry-After` bounds it without trusting a worker count. As the description says, the current bound is load-bearing on production staying at four workers, so a cluster policy change silently invalidates it.

## The 408 per five minute bound does not fit the recorded burst at 25 items

408 request starts per rolling five minutes is 81.6 a minute. The burst on record is about 340,000 rows over thirty minutes, so 11,333 rows a minute. At 25 items a request that needs 453 request starts a minute, which is 5.6x over the bound. At 200 it needs 57, which fits with headroom.

If that holds, the delay does not throttle the burst so much as guarantee the macro flow falls behind during it. Worth checking against your own numbers, because these are exploded macro rows rather than raw rows and that multiplier moves the result.

## Chunking by item count alone is what makes 25 hard to raise

`MAX_ITEM_COUNT` is 1000 in all three sandbox handlers, and `AwsLambdaService` refuses a gzipped response above 50 MB, so what really bounds a chunk is bytes rather than request count. The 25 itself came from #1350, where uncompressed responses of 25 to 50 MB were truncated at Lambda's 6 MB synchronous cap; the `gzip+base64` added in that same PR is the half that fixed it, and the drop from 200 to 25 was never revisited.

Chunking on cumulative serialized size as well as count is what lets the size move at all. I have that on the overhaul branch and am happy to hand it over, or to rebase onto this once it lands.

petar-INFO (approved):

PR #2115, opened 2026-09-21
chore(devkit): enforce the ticket standard with linear:check and linear:create

petar-INFO:

- [ ] Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Test Update
- [x] Chore
- [ ] Other (please describe)

## Description

The Linear standard defined what a ticket must contain; this makes it machine-checkable and lets a reviewed draft become real tickets. A draft is one Markdown file holding one or more tickets with front matter naming the project, team, and state.

`pnpm linear:check` runs the mechanical half of the standard over a draft: the ticket shape, the character budget, bullet length, banned dashes, the title format, and the gate, reporting per ticket rather than failing on the first problem. `pnpm linear:create` turns that draft into Linear issues, and refuses one that does not pass the check. It resolves team, project, state, and labels by name, dry-runs by default so a run prints the plan before anything is written, and on apply creates the tickets and then their comments and relations. Every step is recorded in a sidecar next to the draft, so a run that stops halfway resumes instead of duplicating. The libraries behind them (a ticket-standard parser and a draft parser) are shared, and the Linear skills and the standard doc are updated to point at the new flow.


dominikvrbic (approved):
Lgtm

PR #2130, opened 2026-09-22
perf(data): split macro execution from ingestion and rebuild the gold bridges as streaming CDC tables

petar-INFO:

- [x] Feature
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation Update
- [ ] Performance Improvement
- [ ] Test Update
- [ ] Chore
- [ ] Other (please describe)

## Description

Macro execution ran inside the ingestion pipeline, so a slow macro batch made ingestion queue behind it, and the gold serving tables were shaped for correctness rather than read speed at scale. This restructures the pipeline on both counts.

Macro execution moves into its own pipeline, triggered after Centrum on the dev scheduler, so ingestion no longer waits on it, and macro requests are chunked by serialized size rather than count so a batch stays within limits. The gold bridges are rebuilt as streaming tables fed by AUTO CDC, with the aggregates split out from the lookups above them, and the serving tables are liquid-clustered on `experiment_id` so a per-experiment read touches far less. Along the way the row tracking is dropped and the comments that assumed serverless are corrected, and the Centrum pipeline's event log is published to Unity Catalog so its runs are queryable.

## Linear issues

- Contributes to OJD-1920
- Contributes to OJD-1923
- Contributes to OJD-1924
- Contributes to OJD-1928
- Contributes to OJD-1958


dominikvrbic (approved):

dominikvrbic (dismissed):
> **AI-drafted review**, posted at the user's request. Findings were consolidated from AI reviewers and checked against the code and local reproductions.

Requesting changes on `33f713b92dbffceb8cb9f17261c1a0f72f520461` for the four findings below.

The original review covered `42cdaba9b`. I checked the subsequent three commits before posting. The explicit stream cutoff and architecture documentation updates address the original unbounded-replay and stale-documentation complaints, so those are not repeated here. The remaining affected code is unchanged.

### P1: Bootstrap the macro fact table before switching the Centrum reader

[enriched_experiment_macro_data.py:46–47](https://github.com/Jan-IngenHousz-Institute/open-jii/blob/33f713b92dbffceb8cb9f17261c1a0f72f520461/apps/data/src/pipelines/centrum/enriched/enriched_experiment_macro_data.py#L46)

On the first deployment, `centrum.fact_macro_result` does not exist until the new macro pipeline creates it. Centrum now reads that table unconditionally, but the dev scheduler only starts the macro task after Centrum succeeds. Centrum therefore fails resolving the missing table, and the task that would create it cannot run. Terraform's module dependency creates pipeline definitions, not their output tables.

Add a staged bootstrap before switching the reader, or change the pipeline ownership/order to remove this dependency cycle. Fresh environments also need a path that does not require a legacy macro table to exist.

### P1: Migrate the metrics consumers off the frozen legacy table

[platform_totals.py:51](https://github.com/Jan-IngenHousz-Institute/open-jii/blob/33f713b92dbffceb8cb9f17261c1a0f72f520461/apps/data/src/pipelines/metrics/platform_totals.py#L51), plus `daily_activity.py:53`, `daily_activity_by_resource.py:58`, and `parameter_stats.py:51` in the same directory.

All four still read `EXPERIMENT_MACRO_DATA_TABLE`, which resolves to `experiment_macro_data`. This PR removes its producing notebook from both Centrum pipelines, while new results go to `fact_macro_result`. The metrics jobs will continue succeeding against stale history: totals omit new executions, daily/resource activity misses them, and macro-derived parameter statistics disappear as old measurements leave the **90-day** window.

Switch these consumers to the new fact table in the same migration, after establishing its complete history.

### P2: Bound the Lambda payload after workbook context restoration

[backend_client.py:298–301](https://github.com/Jan-IngenHousz-Institute/open-jii/blob/33f713b92dbffceb8cb9f17261c1a0f72f520461/apps/data/src/lib/enrich/enrich/backend_client.py#L298)

The HTTP byte calculation matches the transmitted request, but it does not bound the Lambda invocation. The backend restores each `$macroInput` context reference into another copy of the measurement before serialization, and adds the macro script.

Using the actual Python chunker and TypeScript normalization/restoration helpers, two 1.7 MB inputs with one context marker each remained one **3,400,340-byte HTTP request**, then expanded into a **6,800,184-byte Lambda payload**. That exceeds the synchronous **6,291,456-byte** limit. Each item fits individually, but the combined invocation fails both. A second reviewer independently reproduced the same failure with 20 smaller inputs. [AWS Lambda quotas](https://docs.aws.amazon.com/lambda/latest/dg/gettingstarted-limits.html).

Split the final per-macro payload after context restoration, including script overhead, or calculate an equivalent safe bound. This is an incomplete fix to the existing payload-size failure.

### P2: Decouple dev ingestion scheduling from macro completion

[dev/main.tf:816–820](https://github.com/Jan-IngenHousz-Institute/open-jii/blob/33f713b92dbffceb8cb9f17261c1a0f72f520461/infrastructure/env/dev/main.tf#L816)

The macro task is a blocking task in the same scheduled job as Centrum, which retains `max_concurrent_runs = 1` at line 792. A macro update taking hours keeps that job active, preventing subsequent half-hour ingestion runs from starting. Separate Spark compute removes slot contention but retains the delay through job concurrency.

Give ingestion an independent schedule and macros their own concurrency control. This finding concerns dev; prod uses continuous pipelines. [Databricks job concurrency](https://docs.databricks.com/aws/en/jobs/configure-job).

### Validation and limits

- On `42cdaba9b`: 136 non-Spark tests and 13 Spark tests passed across the reviewers; Ruff lint and formatting checks passed. These do not exercise Databricks migration or scheduling.
- On `33f713b92`: inspected the full intervening diff and ran the two new cutover tests, both passed. The entire suite was not rerun on this commit.
- Payload overflow reproduced locally without HTTP or Lambda invocation.
- No live Databricks update or migration was executed. Linked Linear acceptance criteria were unavailable.
- The new timestamp cutoff still needs a migration procedure tied to what the old macro flow actually processed, including any raw backlog. A stop-time timestamp alone does not prove that the two flows cover all measurements exactly once.

dominikvrbic (approved):

